Issue #46, opened 2022-07-05
Install MUI in React Native


theblockstalk:
Done

- [ ] https://reactnativepaper.com/ installed in react native and a test component renders

PR #56, opened 2022-07-14
Feature/46 add mui

Rebal67:
issue #46


README.md
```diff
@@ -32,3 +32,10 @@ Features:
   - Android: [Client Android](https://play.google.com/store/apps/details?id=host.exp.exponent&referrer=blankexample)
   - Web: Any web browser
 
+## Contributing
+
+- For folder structure, see [article](https://learn.habilelabs.io/best-folder-structure-for-react-native-project-a46405bdba7)
+- For persistent data storage check:
```
theblockstalk, line 38: i think untill we use these, not mention in README.md

src/components/tmy-button.tsx
```diff
@@ -0,0 +1,12 @@
+import { Children } from "react";
+import { Button } from "react-native-paper";
+
+type Props = {
```
theblockstalk, line 4: can we not do something simpler like

```js
import { Button } from "react-native-paper";
export default Button as TButton
```
?

theblockstalk: also, can we do TButton instead of TMYButton? saves typing two capital chars for every component

theblockstalk (commented):
Finished review

babel.config.js
```diff
@@ -1,6 +1,7 @@
-module.exports = function(api) {
+module.exports = function (api) {
   api.cache(true);
   return {
     presets: ['babel-preset-expo'],
+    plugins: ['react-native-paper/babel'],
```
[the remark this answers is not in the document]
Rebal67: this plugin is for bundling the library to only import the used components instead of the whole library which is more efficient for production

README.md
Rebal67: i will remove them

Issue #91, opened 2022-08-31
Install a global state management tool


theblockstalk:
Definition of done

- [ ] User store can be accessed from any component in the RN app
- [ ] username of the logged in user is available globally
- [ ] function that returns if someone is logged in is available globally

**Hints**

Two options:

- React hooks
- Redux
- maybe another one @Rebal67 knows about

Rebal67:
https://www.reactnativeschool.com/comparing-redux-zustand-and-mobx-state-tree

Zustand for simplicity

PR #98, opened 2022-09-03
added zustand

Rebal67:
issue #91


src/store/userStore.ts
```diff
@@ -0,0 +1,23 @@
+import create from 'zustand';
```
theblockstalk, line 1: This needs to be installed in package.json

src/store/userStore.ts
```diff
@@ -0,0 +1,23 @@
...
+
+
+interface UserState {
+  username: string | null;
```
theblockstalk, line 5: how about we put the User class instance in here?

When the state is created, it can create a new User class. then this is available everywhere easily

theblockstalk (commented):
also check your line spaces, several are unnecessary

src/store/userStore.ts
theblockstalk: add on new line above this
```
// TODO change this to be an instance of User class when we have implemented the RNKeyStore
```

theblockstalk (approved):

PR #105, opened 2022-09-07
Feature/83 create account UI beta Flexbox

Riae:
I can't seem to figure out how to make Flexbox work with RN Paper. I've scoured the internet for a bit, but there's nothing on this. Any ideas?


theblockstalk:
PR is to master, should be to development

PR #176, opened 2022-10-20
Using userStore to get user and save username and password

theblockstalk:


src/utils/RNKeyManager.ts
```diff
@@ -31,34 +44,38 @@ export default class RNKeyManager implements KeyManager {
...
+    };
     if (options.level === KeyManagerLevel.PASSWORD || options.level === KeyManagerLevel.PIN) {
-      if (!options.challenge) throw new Error("Challenge missing");
+      if (!options.challenge) throw new Error('Challenge missing');
       keyStore.salt = randomString(32);
       keyStore.hashedSaltedChallenge = sha256(options.challenge + keyStore.salt);
     }
-    await SecureStore.setItemAsync(options.level, JSON.stringify(keyStore), { requireAuthentication: true });
+
+    const requireAuthentication = options.level === KeyManagerLevel.FINGERPRINT ? true : false;
```
theblockstalk, line 64: this was the main change in this file

src/utils/storage.ts
```diff
@@ -7,16 +7,15 @@ export default class Storage implements PersistantStorage {
 
   async retrieve(key: string): Promise<any> {
     const data = await AsyncStorage.getItem(key);
-    return data;
+    return JSON.parse(data);
```
theblockstalk, line 10: and fixed to ensure always string is handled in storage

PR #205, opened 2022-11-02
Feature/186 username

theblockstalk:


theblockstalk:
getting stuck on keyboard overlay. looking at
https://www.freecodecamp.org/news/how-to-make-your-react-native-app-respond-gracefully-when-the-keyboard-pops-up-7442c1535580/

theblockstalk:
and
https://reactnative.dev/docs/dimensions

src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+        height: `20%`,
+    },
+    username: {
+        height: '20%',
+    },
+    button: {
+        marginTop: 24,
+        marginBottom: 16,
+    },
+    centeredText: {
```
Rebal67, line 115: the styles can be minified by using flex box 
example in this situation:
const layoutStyles = StyleSheet.create({
    container: {
        flex: 1,
        flexDirection: 'column',
        justifyContent: 'space-between',
        paddingHorizontal: 20,
        paddingVertical: 40,
    },
    title: {},
    body: {},
    bottom: {},
});
and removing the all other properties will fix the problem and make everything responsive

src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+                    <TButton onPress={onNext} disabled={username.length === 0} loading={loading}>
+                        Next
+                    </TButton>
+                </View>
+                <View style={styles.centeredText}>
+                    <Text style={styles.bottomMessage}>
+                        Already have an account? <TLink href="login">Login</TLink>
+                    </Text>
+                </View>
+            </View>
```
Rebal67, line 83: too many views added which doesn't really need to be added 
only 2  views are needed 1 for head and 1 for footer this will make everything more efficient

src/containers/HomeScreenContainer.tsx
```diff
@@ -8,7 +8,7 @@ export default function HomeScreenContainer({ navigation }: { navigation: Naviga
     return (
         <View>
             <View>
-                <TButton onPress={() => navigation.navigate('createAccount')}>Create Account</TButton>
+                <TButton onPress={() => navigation.navigate('createAccountUsername')}>Create Account</TButton>
                 <TButton onPress={() => navigation.navigate('login')}>Login</TButton>
             </View>
```
Rebal67, line 14: adding margin to the top and bottom really help as some phones don't have a notch the first 35 on top of screens aren't touchable use paddingTop 40 just to be sure
or  paddingVertical: 40; for consistency

src/navigation/main.tsx
```diff
@@ -3,32 +3,68 @@ import { createNativeStackNavigator } from '@react-navigation/native-stack';
...
+        backgroundColor: theme.colors.primary,
+        height: 164,
+    },
+    headerTitleStyle: {
+        fontSize: 24,
+    },
+    headerTitleAlign: 'center',
+    headerTintColor: 'white',
+};
+
```
Rebal67, line 29: can we maybe move this into theme.ts

src/navigation/main.tsx
```diff
@@ -3,32 +3,68 @@ import { createNativeStackNavigator } from '@react-navigation/native-stack';
...
+                            name="home"
+                            options={{ ...defaultScreenOptions, headerShown: false }}
+                            component={HomeScreen}
+                        />
+                        <Stack.Screen
+                            name="securitySplash"
+                            options={{ ...defaultScreenOptions, headerShown: false }}
+                            component={SplashSecurityScreen}
+                        />
+                        <Stack.Screen
```
Rebal67, line 52: is it necessary to add defaultscreen by each  screen?

src/theme.ts
```diff
@@ -11,6 +11,7 @@ const theme: Theme = {
         primary: settings.config.theme.primaryColor,
         accent: settings.config.theme.secondaryColor,
         surface: settings.config.theme.tertiaryColor,
+        background: 'white',
     },
 };
```
Rebal67, line 17: move this file to utils ?

src/components/TUsername.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export default function TUsername(props: TTextInputProps) {
+  const showError: boolean = !!props.errorText && props.errorText.length > 0;
+
+  return (
+    <View>
+      <View style={styles.username}>
+        <TextInput style={styles.usernameInput} {...props} error={showError} />
+        <Text style={styles.accountSuffix}>{props.suffix}</Text>
+      </View>
```
Rebal67, line 15: too many views
why not use TTextInput  and TText ?

src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+    },
+});
+
+const styles = StyleSheet.create({
+    message: {
+        height: `20%`,
+    },
+    username: {
+        height: '20%',
+    },
```
Rebal67, line 110: avoid using height if possible

src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    const user = useUserStore().user;
+
+    async function onNext() {
+        setLoading(true);
+
+        try {
+            await user.saveUsername(username, settings.config.accountSuffix);
+        } catch (e) {
+            if (e instanceof ExpectedSdkError && e.code === 'TSDK1000') {
```
Rebal67, line 30: how can we make the error codes more readable for developers ?

Rebal67 (changes requested):

src/containers/CreateAccountUsernameContainer.tsx
theblockstalk: hmmm, I'm not sure about this. it spaces everything evenly through the screen vertically which is not the design

theblockstalk: especially if we want the NEXT button to be in the same place on all/most screens

src/components/TUsername.tsx
theblockstalk: I use all the View to ensure the error message goes underneath both the input and the suffix

src/containers/CreateAccountUsernameContainer.tsx
theblockstalk: hmmm yes maybe a good idea. We could use an enum, but this should be another task. e.g.
`e.code === SDKErrors.UsernameExists`

src/containers/CreateAccountUsernameContainer.tsx
Rebal67: we can add flex property to the head and footer flex: 4 for the head and flex: 1 for the footer that will keep the lower part always starts at the same level

PR #250, opened 2022-11-21
Feature/151 create pin UI

Riae:


src/components/TPin.tsx
```diff
@@ -0,0 +1,126 @@
...
+        setPin(function (pin) {
+            if (pin.length === 5) {
+                return pin;
+            } else {
+                return pin + num;
+            }
+        });
+    }
+    function DotOrChar(props: any) {
+        if (props.pin.length >= props.pinIndex + 1) {
```
theblockstalk, line 21: this is still not a modular component, so I will not accept this. it should accept the least needed in this case just the char to display or not a char

src/components/TPin.tsx
```diff
@@ -0,0 +1,126 @@
...
+                </TouchableOpacity>
+            </View>
+        </>
+    );
+}
+
+const styles = StyleSheet.create({
+    text: {
+        fontSize: 35,
+        // Color: '#000000',
```
theblockstalk, line 75: remove comments

src/containers/PinScreenContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const handlePin = (p: string) => {
+        setPin(p);
+    };
+    return (
+        <View style={styles.head}>
+            <Text style={styles.header}>
+                <TH1>Add a PIN</TH1>
+            </Text>
+            <Text style={styles.headdescription}>This helps keep your account secure</Text>
+            <TPin onPin={handlePin}></TPin>
```
theblockstalk, line 21: `onPin={setPin}` is better

theblockstalk (changes requested):
very good! just one small commend and a component to fix

src/containers/PinScreenContainer.tsx
Riae: It conflicts with the SetPin function in TPin. change it?

PR #255, opened 2022-11-22
Feature/194 menu UI

Rebal67:


Rebal67:
coloring will automatically be fixed once the main screen ui get merged

README.md
```diff
@@ -26,25 +26,24 @@ Features:
...
+      "eas": {
         "projectId": "afffe2ee-9f93-4d18-9361-df30429cbd98"
+      }
     }
-}
 ```
 
-For future runs, make sure the `projectId` matches the one in [https://expo.dev/](https://expo.dev/) website.
-
-5. (optional for IOS only) Run `npm run build:ios:create` to create a device profile for your phone (you need to connect to an Apple Developer account - we do not have one yet)
```
theblockstalk, line 44: we should use these commands with npm, makes it easier for developer

theblockstalk: I think these instructions are more up to date

package.json
```diff
@@ -6,7 +6,7 @@
     "build:ios:create": "eas device:create",
     "build:ios": "npm run build:prepare && eas build --clear-cache --profile development --platform ios",
     "build:android": "npm run build:prepare && eas build --clear-cache --profile development --platform android",
-    "start": "npm run build:prepare && expo start --dev-client --clear",
+    "start": "npm run build:prepare && expo start --dev-client --clear -c",
```
theblockstalk, line 9: --clear and -c are the same, leave --clear

prepare.ts
```diff
@@ -2,6 +2,7 @@ import fs from 'fs';
 import settings from './src/settings';
 import app from './app.default.json';
 import myPackage from './package.json';
+import appPersonal from './app.json';
```
theblockstalk, line 5: what is this for?

src/components/customDrawer.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
theblockstalk, line 1: capital C CustomDrawer.tsx for naming convention

src/screens/MenuScreen.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
theblockstalk, line 1: where is this file used?

src/settings.ts
```diff
@@ -10,9 +12,9 @@ type ConfigType = {
     ecosystemName: string;
     appSlogan: string;
     images: {
-        splash: string;
-        logo48: string;
-        logo1024: string;
+        splash: ImageSourcePropType;
```
theblockstalk, line 15: why do this? leave them as URLs and let the component do with it what they need. the config object should have the raw inputs (string URLs), not the objects they point to. this can be put in the Settings object, which contains the information derived from the Config object

src/settings.ts
```diff
@@ -42,7 +44,12 @@ const settings: SettingsType = {
 switch (env) {
     case 'development':
         config = require('./config/config.json');
+        // TODO find a better way switch images
+        config.images.logo1024 = require('./assets/tonomy/tonomy-logo1024.png');
```
theblockstalk, line 48: this should not be here. if anything, it can be in Settings object, should not use this string literal, should use the value of config.images.logo1024 string

src/containers/HomeScreenContainer.tsx
```diff
@@ -23,7 +23,7 @@ export default function HomeScreenContainer({ navigation }: { navigation: Naviga
                     >
                         Create Account
                     </TButton>
-                    <TButton mode="contained" onPress={() => navigation.navigate('test')}>
+                    <TButton mode="contained" onPress={() => navigation.navigate('main')}>
```
theblockstalk, line 26: this is temporary, to test it out, right?

index.js
```diff
@@ -1,3 +1,5 @@
+import 'react-native-gesture-handler';
```
theblockstalk, line 1: this is not used anywhere. why is it here?

src/navigation/drawer.tsx
```diff
@@ -0,0 +1,20 @@
...
+import CustomDrawer from '../components/customDrawer';
+import CreateAccountUsernameScreen from '../screens/CreateAccountUsernameScreen';
+
+const Drawer = createDrawerNavigator();
+
+export default function DrawerNavigation() {
+    return (
+        <Drawer.Navigator drawerContent={(props) => <CustomDrawer {...props} />} initialRouteName="t">
+            <Drawer.Screen name="t" component={HomeScreen} />
+            <Drawer.Screen name="settings" component={HomeScreen} />
```
theblockstalk, line 15: shouldn't the rest of these also have titles? like for help

index.js
Rebal67: it is used by the app automatically thhe app will crash without it

src/navigation/drawer.tsx
Rebal67: they have the same title as the  name

README.md
Rebal67: these instruction are meant for expo without EAS and people struggle with running the design only 
to make them run with npm please make a new task

package.json
Rebal67: removed

src/components/customDrawer.tsx
Rebal67: done

src/settings.ts
Rebal67: i was trying to use images inside the app from thhe settings but we need to find another way to do this and it will mostly be via babel

src/containers/HomeScreenContainer.tsx
Rebal67: yes

src/settings.ts
Rebal67: removed

src/screens/MenuScreen.tsx
Rebal67: deleted it was first used for testing

prepare.ts
Rebal67: old settings removed

README.md
theblockstalk: with designonly you still need to run EAS right?

Rebal67: yes

theblockstalk: then we need to keep then, and we should make them the same as on https://github.com/Tonomy-Foundation/Tonomy-ID-Integration/tree/development#pre-run-one-time

Rebal67: done

theblockstalk: see steps here. they are different. please align them
https://github.com/Tonomy-Foundation/Tonomy-ID-Integration/tree/development#pre-run-one-time

Issue #196, opened 2022-10-27
Github action - Tonomy ID unit tests


theblockstalk:
Definition of done

- [x] unit tests are run in Github action on every commit

Follow up: linting also run

PR #271, opened 2022-11-28
Issue #196 GitHub actions workflow to run tests on every pull request

techemmy:


.github/workflows/build-tests.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Use Node.js 16.4.1
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.4.1
+      - run: npm install
+      - run: npm run build --if-present
```
theblockstalk, line 18: is this step necessary - I think it might not be? can you remove and see if it works?

techemmy: I think line `17` is needed to install the dependencies, but 18 isn't.
Let me edit it and see if it works

PR #280, opened 2022-12-05
Feature/263 error handler

theblockstalk:


App.tsx
```diff
@@ -7,12 +7,21 @@ import MainNavigation from './src/navigation/main';
...
+
+    setErrorHandlers(errorStore);
+
     return (
         <PaperProvider theme={theme}>
             <SafeAreaProvider>
-                <MainNavigation></MainNavigation>
+                <ErrorHandlerContainer>
+                    <MainNavigation />
+                </ErrorHandlerContainer>
```
Rebal67, line 24: i dont think this should parent the whole app as it reloads everything when state is changed 
it's better to put it at same level with main Navigation

src/components/ErrorHandlerProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+            useErrorStore.subscribe((state) => {
+                errorRef.current.error = state.error;
+                errorRef.current.title = state.title;
+                errorRef.current.expected = state.expected;
+                if (state.error) {
+                    setShowModal(true);
+                }
+            }),
+        []
+    );
```
Rebal67, line 33: in theory zustand reloads the component when the state is changed. what is the reason that made it not do that behavior.
why use RefHook ?

src/components/TErrorModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+            return (
+                <View>
+                    <TP size={2}>Path: {props.error.path}</TP>
+                    <TP size={2}>Response: {props.error.response}</TP>
+                    <TP size={2}>SourceUrl: {props.error.sourceURL}</TP>
+                </View>
+            );
+        } else if (props.error instanceof EosioUtil.AntelopePushTransactionError) {
+            const trxError = props.error.error;
+
```
Rebal67, line 45: shouldn't this only be for developers ?

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -140,20 +145,19 @@ export default function CreateAccountPasswordContainer({ navigation }: { navigat
...
+                expected={true}
             >
                 <View>
                     <Text>
                         Username <Text style={{ color: theme.colors.primary }}>{user.storage.username.username}</Text>{' '}
                         is already taken. Please choose another one.
                     </Text>
                 </View>
-            </TModal>
+            </TErrorModal>
```
Rebal67, line 160: can't we remove this? and make the root component show ?

Rebal67: maybe we can add this for all kind of dialogs

src/components/ErrorHandlerProvider.tsx
theblockstalk: in practice, a change in useErrorStore() state did not trigger a re-render. A re-render is only triggered on state (useState) or props change. This was the workaround I found to this

src/containers/CreateAccountPasswordContainer.tsx
theblockstalk: this has a special logic, which is that it navigates the user back to the username screen instead of just closing. because of this it's implemented here.

PR #325, opened 2022-12-28
Feature/222 sso login screen logic

theblockstalk:


src/containers/FingerprintUpdateContainer.tsx
```diff
@@ -1,23 +1,22 @@
...
-import { Props } from '../screens/FingerprintUpdateScreen';
+import { useRootNavigator } from '../navigation/Root';
 
 export default function CreateAccountContainer({ password }: { password: string }) {
     const [showModal, setShowModal] = useState(false);
     const user = useUserStore((state) => state.user);
     const errorStore = useErrorStore();
 
-    const navigation = useNavigation<Props['navigation']>();
+    const navigation = useRootNavigator();
```
Rebal67, line 20: this will not differ from using useNavigation<any> specifying the page u are in will help catch wrong navigations at runtime although it's much work it's worth it when changing routes

src/screens/FingerprintUpdateScreen.tsx
```diff
@@ -3,8 +3,6 @@ import React from 'react';
 import FingerprintUpdateContainer from '../containers/FingerprintUpdateContainer';
 import { RouteStackParamList } from '../navigation/Root';
 
-export type Props = NativeStackScreenProps<RouteStackParamList, 'CreateAccountFingerprint'>;
-
```
Rebal67, line 7: this should be exported in each screen with the right naming

PR #353, opened 2023-01-05
Feature/341 design changes implemented

StefanoPitto:


src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -99,3 +100,17 @@ export default function CreateAccountUsernameContainer({ navigation }: Props) {
...
+
+const styles = StyleSheet.create({
+    caption: {
+        textAlign: 'right',
+    },
+    userInput: {
+        backgroundColor: 'transparent',
+        flex: 1,
+    },
+    inputContainer: {
```
theblockstalk, line 112: Design is going to have the boarder for ALL form components. So better we should put this in TUsername component.

This border should also be applied to the TPassword and TTextInput components

src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -99,3 +100,17 @@ export default function CreateAccountUsernameContainer({ navigation }: Props) {
...
+        borderWidth: 1,
+        borderColor: '#E0E0E0',
```
theblockstalk, line 114: use theme color disabled

StefanoPitto: Updated.

src/containers/CreateAccountUsernameContainer.tsx
StefanoPitto: Already updated, moved to the TUsername component.

PR #358, opened 2023-01-10
Feature/341 UI changes implemented v2

StefanoPitto:
Needed to create another branch, run 'npm run lint' to fix linter issues. The navigation problem is really complex, it happens because of a Require cycle. Didn't change any of those files before so do not know why it happened. Let me know if there's any other issue.


src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -89,42 +93,64 @@ export default function CreateAccountPasswordContainer({ navigation }: Props) {
...
                 footerHint={
-                    <View style={[commonStyles.marginBottom, commonStyles.alignItemsCenter]}>
-                        <TP size={1} style={[styles.rememberPasswordText, commonStyles.textAlignCenter]}>
-                            Please remember your master password for future use
-                        </TP>
+                    <View style={[commonStyles.marginBottom]}>
+                        <View style={commonStyles.alignItemsCenter}>
+                            <TP size={1}>
+                                By continuing, you agree to our
+                                <TP size={1} style={styles.underline}>
```
theblockstalk, line 134: T&C and Privacy Should both be hyperlinks that when opened, are sent to a URL which is provided in the config files. Use the <TA> Tonomy component, and use a link to a new value in config files (see example here:
[setErrorMessage](https://github.com/Tonomy-Foundation/Tonomy-ID/pull/358/files#diff-53470cebe81107675c830aaebb31ae60069a51f45d8082c67e6f723be869d649R150))

src/settings.ts
```diff
@@ -17,6 +17,8 @@ type ConfigType = {
         logo1024: string;
     };
     links: {
+        privacyPolicy: string;
```
theblockstalk, line 20: you also need to create the corresponding values in config.json and config.staging.json "#" like the other links

PR #359, opened 2023-01-10
add app

Rebal67:


src/containers/SSOLoginContainer.tsx
```diff
@@ -44,8 +45,8 @@ export default function SSOLoginContainer({ requests }: { requests: string }) {
                 const payload = jwt.payload as JWTLoginPayload;
                 if (payload.origin === settings.config.ssoWebsiteOrigin) {
                     setTonomyIdJwtPayload(payload);
-                    // TODO next line, but need to add this app to Bootstrap script first
-                    // const app = await App.getApp(payload.origin);
+                    const app = await App.getApp(payload.origin);
```
theblockstalk, line 48: L48 and L52 can both put before the if/else condition logic, after L45

src/containers/SSOLoginContainer.tsx
```diff
@@ -63,14 +64,16 @@ export default function SSOLoginContainer({ requests }: { requests: string }) {
 
     async function onNext() {
         try {
-            await user.apps.loginWithApp(ssoApp, ssoJwtPayload?.publicKey);
-
             let callbackUrl = settings.config.ssoWebsiteOrigin + '/callback?';
             callbackUrl += 'requests=' + requests;
```
theblockstalk, line 68: If the `requests` parameter is not uused at all then maybe we can not send it back in the URL... What do you think?

Rebal67: i think we might need it let's keep it for now

PR #363, opened 2023-01-16
Feature/152 login UI implemented

StefanoPitto:


src/containers/LoginPasswordContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+            }
+            footer={
+                <View>
+                    <View style={commonStyles.marginBottom}>
+                        <TButtonContained>NEXT</TButtonContained>
+                    </View>
+                    <View style={commonStyles.alignItemsCenter}>
+                        <TP size={1}>
+                            Don&apos;t have an account? <TLink href="signup">Sign up</TLink>
+                        </TP>
```
Rebal67, line 55: u can use ' instead of codes because TP component will see it as a string anyway
please use navigation instead of href

src/containers/LoginPasswordContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+                </View>
+            }
+        ></LayoutComponent>
+    );
+}
+const styles = StyleSheet.create({
+    container: {
+        width: '100%',
+        marginVertical: '33%',
+    },
```
Rebal67, line 66: dont use margin's and paddings for positioning and the layout those are supposed to be used for white spacing between components;
use FLex and grid instead;

example: 
flex:1;
alignItems: "center";

src/containers/LoginUsernameContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+                            }}
+                        >
+                            NEXT
+                        </TButtonContained>
+                    </View>
+                    <View style={commonStyles.alignItemsCenter}>
+                        <TP size={1}>
+                            Don&apos;t have an account? <TLink href="signup">Sign up</TLink>
+                        </TP>
+                    </View>
```
Rebal67, line 63: same as review above

src/containers/LoginUsernameContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+const styles = StyleSheet.create({
+    inputContainer: {
+        borderWidth: 1,
+        borderColor: theme.colors.disabled,
+    },
+    container: {
+        width: '100%',
+        marginVertical: '33%',
```
Rebal67, line 77: same here

Rebal67 (changes requested):

Rebal67 (approved):

Issue #117, opened 2022-09-14
Password generator


theblockstalk:
A random and secure password is generated for the user in the form, instead of them choosing one.

Definition of done:

- [ ] Only one password form field exists by default
- [ ] the password field is prefilled with a cryptographically secure randomly generated password that meets the password validation requirements
- [ ] if the user changes the generated password, then a second form field appears where they need to confirm the password they create

theblockstalk:
invalid as we are no longer using password

PR #394, opened 2023-01-20
#117: Password generator

MZO9400:
Automatically generate passwords based off of Node runtime's crypto module.

[X]  Only one password form field exists by default 
[X]  the password field is prefilled with a cryptographically secure randomly generated password that meets the password validation requirements 
[X]  if the user changes the generated password, then a second form field appears where they need to confirm the password they create


MZO9400:
This PR fixes #117

PR #395, opened 2023-01-23
Feature/73 login logic implemented

StefanoPitto:


src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -23,7 +23,6 @@ export default function CreateAccountUsernameContainer({ navigation }: Props) {
     const [errorMessage, setErrorMessage] = useState('');
     const [loading, setLoading] = useState(false);
 
-    const user = useUserStore().user;
```
theblockstalk, line 26: this is needed for L33

PR #401, opened 2023-01-24
Feature/pin UI for Login Process

HaseebUllahAbbasi:
Pin Screen for the Login Process which checks the user Pin if corrects the proceed to next screen otherwise shows message

### DIFF
--- a/src/containers/LoginPinScreenContainer.tsx
+++ b/src/containers/LoginPinScreenContainer.tsx
@@ -1,0 +1,94 @@
+import React, { useState } from 'react';
+import { StyleSheet, Text, View } from 'react-native';
+import TButton from '../components/atoms/Tbutton';
+import { TH1 } from '../components/atoms/THeadings';
+import TPin from '../components/TPin';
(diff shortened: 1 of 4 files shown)


src/containers/LoginPinScreenContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+import TButton from '../components/atoms/Tbutton';
+import { TH1 } from '../components/atoms/THeadings';
+import TPin from '../components/TPin';
+import useUserStore from '../store/userStore';
+import { HelperText } from 'react-native-paper';
+import LayoutComponent from '../components/layout';
+import useErrorStore from '../store/errorStore';
+import { Props } from '../screens/FingerprintUpdateScreen';
+import theme from '../utils/theme';
+export default function LoginPinScreenContainer({
```
theblockstalk, line 12: a lot of this is duplicated from the Create AccountScreenContainer.

Can you please try to use the same container for both and move some of the logic to the Login / CreateAccountScreen component instead?

PR #414, opened 2023-01-25
Readme/prebuild steps changes

HaseebUllahAbbasi:
Read me Changes for the Pre Build Steps

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ npm i
 
 ## Pre-run build (first time and each time new RN only packages are installed)
 
-This is to create an expo build so you can down an `.apk` or `.ipa` file from [https://expo.dev](https://expo.dev) which you can use to run the app.
+This is to create an expo build so you can down an `.apk` or `.ipa` file from [https://expo.dev](https://expo.dev) which you can use to run the app. You can't Run from the Official Expo App.  
 
 1. Create an expo account to build the app. [https://expo.dev/signup](https://expo.dev/signup)
 2. (for first time build only) Remove the following lines from `app.default.json`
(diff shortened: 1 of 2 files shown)


src/containers/LoginPinScreenContainer.tsx
```diff
@@ -71,6 +73,9 @@ export default function LoginPinScreenContainer({
                     >
                         {'Next'}
                     </TButton>
+                    <TButtonOutlined onPress={onSkip} style={styles.marginBottom}>
+                        Skip
+                    </TButtonOutlined>
```
Rebal67, line 78: this edit should be in different PR, but for now it's ok.
for next time we handle 1 branch per issue

Rebal67 (commented):
for future PRs we have naming convention which is:
feature/ISSUENUMBER-some-description.
and for hotfixes: 
hotfix/some-description

src/containers/LoginPinScreenContainer.tsx
HaseebUllahAbbasi: > 

It was just readme changes so i added as readme

README.md
```diff
@@ -37,8 +37,32 @@ This is to create an expo build so you can down an `.apk` or `.ipa` file from [h
 3. (to connect to an existing build in [https://expo.dev](https://expo.dev)) change the value of "projectId" in `app.default.json` to the vale of the project in [https://expo.dev](https://expo.dev)
```
theblockstalk, line 37: make it clear that they should to step 2 or 3 but not both

README.md
```diff
@@ -21,7 +21,7 @@ npm i
 
 ## Pre-run build (first time and each time new RN only packages are installed)
 
-This is to create an expo build so you can down an `.apk` or `.ipa` file from [https://expo.dev](https://expo.dev) which you can use to run the app.
+This is to create an expo build so you can down an `.apk` or `.ipa` file from [https://expo.dev](https://expo.dev) which you can use to run the app. You can't Run from the Official Expo App.  
 
 1. Create an expo account to build the app. [https://expo.dev/signup](https://expo.dev/signup)
 2. (for first time build only) Remove the following lines from `app.default.json`
```
theblockstalk, line 27: 2. (for first time build only) Remove the following lines from `app.default.json`
```json
    "extra": {
      "eas": {
        "projectId": "afffe2ee-9f93-4d18-9361-df30429cbd98"
      }
    }
```
Replace with
```json
    "extra": {
      "eas": {}
    }
```

README.md
```diff
@@ -37,8 +37,32 @@ This is to create an expo build so you can down an `.apk` or `.ipa` file from [h
...
+```json
+    "extra": {
+      "eas": {
+      }
+    }
+```
+
+and go the prepare.ts and add this before  the ' if ' condition ends
+
+```typeScript
```
theblockstalk, line 57: we should fix this problem in the prepare.ts file instead. then we don't need to mention this fix here.

src/containers/LoginPinScreenContainer.tsx
theblockstalk: @Rebal67 is right it should be in a separate PR

PR #423, opened 2023-01-28
Bug/421 password validation UI

HaseebUllahAbbasi:
Now user can see what is password status how strong is it as it will be approved or not

### DIFF
--- a/src/containers/CreateAccountPasswordContainer.tsx
+++ b/src/containers/CreateAccountPasswordContainer.tsx
@@ -18,8 +18,8 @@ import { Props } from '../screens/CreateAccountPasswordScreen';
 import TA from '../components/atoms/TA';
 
 export default function CreateAccountPasswordContainer({ navigation }: Props) {
-    const [password, setPassword] = useState(!settings.isProduction() ? '' : '');
-    const [password2, setPassword2] = useState(!settings.isProduction() ? '' : '');
+    const [password, setPassword] = useState(!settings.isProduction() ? 'k^3dTEqXfolCPo5^QhmD' : '');
+    const [password2, setPassword2] = useState(!settings.isProduction() ? 'k^3dTEqXfolCPo5^QhmD' : '');
     const [errorMessage, setErrorMessage] = useState('');
     const [confirmErrorMessage, setConfirmErrorMessage] = useState('');
     const [loading, setLoading] = useState(false);
--- a/src/containers/CreateAccountPasswordContainer.tsx
+++ b/src/containers/CreateAccountPasswordContainer.tsx
@@ -18,9 +18,11 @@ import { Props } from '../screens/CreateAccountPasswordScreen';
 import TA from '../components/atoms/TA';
 
 export default function CreateAccountPasswordContainer({ navigation }: Props) {
-    const [password, setPassword] = useState(!settings.isProduction() ? 'k^3dTEqXfolCPo5^QhmD' : '');
-    const [password2, setPassword2] = useState(!settings.isProduction() ? 'k^3dTEqXfolCPo5^QhmD' : '');
+    const [password, setPassword] = useState(!settings.isProduction() ? '' : '');
+    const [password2, setPassword2] = useState(!settings.isProduction() ? '' : '');
     const [errorMessage, setErrorMessage] = useState('');
+    const [confirmErrorMessage, setConfirmErrorMessage] = useState('');
+
     const [loading, setLoading] = useState(false);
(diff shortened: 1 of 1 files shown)


HaseebUllahAbbasi:
but there is overlapping issue is still there I will figure it out, and that is not only on one screen then it has to implmented on call the screens when the keyboard opens

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -18,19 +18,37 @@ import { Props } from '../screens/CreateAccountPasswordScreen';
...
     const [showUsernameErrorModal, setShowUsernameErrorModal] = useState(false);
     const [error, setError] = useState(false);
     const user = useUserStore().user;
     const errorStore = useErrorStore();
-
+    const passwordRegex = /^(?=.*\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\w\d\s:])([^\s]){12,}$/;
     useEffect(() => {
-        setError(!(password === password2));
+        if (password.length > 0) {
+            const isValid = passwordRegex.test(password);
```
theblockstalk, line 36: this is already done within `user.savePassword` function and does not need to be done twice. Correct error handling of `user.savePassword` can handle the condition you have here already.

theblockstalk: please use `user.savePassword` here instead.

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -18,19 +18,37 @@ import { Props } from '../screens/CreateAccountPasswordScreen';
...
     const [loading, setLoading] = useState(false);
     const [trxUrl, setTrxUrl] = useState('');
     const [showModal, setShowModal] = useState(false);
     const [showUsernameErrorModal, setShowUsernameErrorModal] = useState(false);
     const [error, setError] = useState(false);
     const user = useUserStore().user;
     const errorStore = useErrorStore();
-
+    const passwordRegex = /^(?=.*\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\w\d\s:])([^\s]){12,}$/;
     useEffect(() => {
```
theblockstalk, line 34: make sure all functions have a line space before and after

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -96,23 +114,31 @@ export default function CreateAccountPasswordContainer({ navigation }: Props) {
...
+                                    style={errorMessage.length > 0 ? errorStyles.labelError : styles.labelText}
+                                >
                                     Password
                                 </TP>
                                 <TPasswordInput
                                     value={password}
                                     onChangeText={setPassword}
                                     errorText={errorMessage}
-                                    outlineColor={error ? theme.colors.error : theme.colors.primary}
+                                    outlineColor={errorMessage.length > 0 ? theme.colors.error : theme.colors.primary}
```
theblockstalk, line 127: better to use the `error` state as it is used for both the modal + validation errors

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -96,23 +114,31 @@ export default function CreateAccountPasswordContainer({ navigation }: Props) {
...
                                     style={commonStyles.marginBottom}
                                 />
                             </View>
                             <View>
-                                <TP style={error ? errorStyles.labelError : styles.labelText}>Confirm Password</TP>
+                                <TP style={confirmErrorMessage.length > 0 ? errorStyles.labelError : styles.labelText}>
```
theblockstalk, line 132: same as L127

src/containers/CreateAccountPasswordContainer.tsx
HaseebUllahAbbasi: removed in my new commit

src/containers/CreateAccountPasswordContainer.tsx
HaseebUllahAbbasi: I used it as there are two fields one for password and one for confirm password so this one is for password

src/containers/CreateAccountPasswordContainer.tsx
HaseebUllahAbbasi: I used it as there are two fields one for password and one for confirm password so this one is for confirm password, it is true when user confirm password doesn't matches the password

HaseebUllahAbbasi: @Rebal17 @theblockstalk

PR #424, opened 2023-01-28
Feature/223 adding login desktop

Rebal67:


src/containers/MainContainer.tsx
```diff
@@ -1,18 +1,31 @@
...
+import QrCodeScanContainer from './QrCodeScanContainer';
+
+import { MainScreenNavigationProp } from '../screens/MainScreen';
 
 export default function MainContainer() {
     const user = useUserStore((state) => state.user);
+    const navigation = useNavigation<MainScreenNavigationProp['navigation']>();
     const [username, setUsername] = useState<TonomyUsername>({});
+    const [qrOpened, setQrOpened] = useState<boolean>(false);
     useEffect(() => {
```
theblockstalk, line 23: please add line space before/after all functions/classes

src/containers/MainContainer.tsx
```diff
@@ -23,15 +36,35 @@ export default function MainContainer() {
...
+                                setQrOpened(true);
+                            }}
+                        >
+                            Scan Qr Code
+                        </TButtonContained>
+                    </View>
+                </View>
+            )}
+
+            {qrOpened && <QrCodeScanContainer onScan={onScan} onClose={onClose} />}
```
theblockstalk, line 67: is this a popup, or new screen? i thought this would navigate to new screen? seems more modular to me

src/containers/QrCodeScanContainer.tsx
```diff
@@ -1,16 +1,15 @@
...
 
-export const width = Dimensions.get('window').width * 0.01;
-export const height = Dimensions.get('window').height * 0.01;
-
-export default function QrCodeScanContainer() {
+export default function QrCodeScanContainer(props: {
+    onClose?: () => void;
+    onScan: (result: BarCodeScannerResult) => void;
+}) {
     const [hasPermission, setHasPermission] = useState(null as null | boolean);
```
theblockstalk, line 13: null as null
is this necessary?

src/containers/QrCodeScanContainer.tsx
```diff
@@ -23,62 +22,58 @@ export default function QrCodeScanContainer() {
...
-                <View>
-                    <TButtonOutlined style={commonStyles.marginBottom}>Cancel</TButtonOutlined>
-                </View>
-            }
-        />
+                    />
+                )}
+                {scanned && <TButtonContained onPress={() => setScanned(false)}> Tap to Scan Again</TButtonContained>}
+            </View>
+            {/* <View
```
theblockstalk, line 46: remove comments if not used

src/containers/SSOLoginContainer.tsx
```diff
@@ -25,7 +32,7 @@ export default function SSOLoginContainer({ requests }: { requests: string }) {
     const [ssoJwtPayload, setSsoJwtPayload] = useState<JWTLoginPayload>();
     const [ssoApp, setSsoApp] = useState<App>();
 
-    const navigation = useRootNavigator();
+    const navigation = useNavigation();
```
theblockstalk, line 35: useRootNavigator has type check built in. useNavigator does not. Why remove this?

src/screens/MainScreen.tsx
```diff
@@ -1,6 +1,9 @@
+import { DrawerScreenProps } from '@react-navigation/drawer';
 import React from 'react';
 import MainContainer from '../containers/MainContainer';
+import { RouteDrawerParamList } from '../navigation/Drawer';
 
+export type MainScreenNavigationProp = DrawerScreenProps<RouteDrawerParamList, 'UserHome'>;
```
theblockstalk, line 6: the useRootNavigator removes the need to have to declare these every time

src/containers/SSOLoginContainer.tsx
Rebal67: A hook is linked to component life cycle, RootNavigation uses the RootNavigation life cycle inside other  components cycle which causes an error to use typing we can use useNavigation<NavigationType>

src/screens/MainScreen.tsx
Rebal67: we can export types not hooks and states

src/containers/MainContainer.tsx
Rebal67: then it will be messy to keep sending data through navigation and clearing navigation

src/containers/QrCodeScanContainer.tsx
Rebal67: for keeping logic outside of components and only in containers but we can move the data sharing and use zustand i dont see that's necessary yet

PR #425, opened 2023-01-30
Feature/408 login add pin logic

HaseebUllahAbbasi:
Saving pin and disabled next until user enter the full confirm pin

### DIFF
--- a/src/containers/PinScreenContainer.tsx
+++ b/src/containers/PinScreenContainer.tsx
@@ -37,9 +37,7 @@ export default function PinScreenContainer({
         if (confirming) {
             if (pin === confirmPin) {
                 try {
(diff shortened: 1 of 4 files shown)


src/containers/PinScreenContainer.tsx
```diff
@@ -36,16 +36,16 @@ export default function PinScreenContainer({
         setLoading(true);
         if (confirming) {
             if (pin === confirmPin) {
-                console.log('saving pin');
                 try {
-                    await user.savePIN(pin);
+                    console.log('Pin is About to be saved');
```
theblockstalk, line 40: remove all console.log statements

HaseebUllahAbbasi: removed

PR #469, opened 2023-02-13
finger print new ui update

HaseebUllahAbbasi:
finger print new ui update and border must be merge before it to get the desired layout from the figma

### DIFF
--- a/src/components/atoms/THeadings.tsx
+++ b/src/components/atoms/THeadings.tsx
@@ -44,10 +44,8 @@ export function TCaption(props: TCaptionProps) {
 const styles = StyleSheet.create({
     h1: {
         fontSize: 30,
-        marginTop: 15,
         marginBottom: 8,
         color: customColors.textBold,
-        textAlign: 'center',
     },
     h2: {
         fontSize: 24,
@@ -60,6 +58,5 @@ const styles = StyleSheet.create({
     },
     s1: {
         fontSize: 16,
-        textAlign: 'center',
     },
 });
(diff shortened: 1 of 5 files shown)


src/components/atoms/THeadings.tsx
```diff
@@ -44,8 +44,10 @@ export function TCaption(props: TCaptionProps) {
 const styles = StyleSheet.create({
     h1: {
         fontSize: 30,
```
Rebal67, line 46: this will update styling for all header are we sure we need to do this ?

src/navigation/Root.tsx
```diff
@@ -70,13 +70,15 @@ export default function RootNavigation() {
...
+            // backgroundColor: theme.colors.,
+            backgroundColor: '#F9F9F9',
         },
         headerTitleStyle: {
             fontSize: 24,
+            color: 'black',
         },
         headerTitleAlign: 'center',
-        headerTintColor: theme.dark ? theme.colors.text : 'white',
+        headerTintColor: theme.dark ? theme.colors.text : 'black',
```
Rebal67, line 81: same goes here

HaseebUllahAbbasi: yes I tested on my phone on both themes white, dark

src/components/atoms/THeadings.tsx
HaseebUllahAbbasi: yes I tested this it worked on both white and dark mode @Rebal67

src/navigation/Root.tsx
HaseebUllahAbbasi: @Rebal67 then how do you suggest then ? it works from on both theme white and dark, on my phone, let me know if you  
have suggestions

src/components/atoms/THeadings.tsx
HaseebUllahAbbasi: @Rebal67  the header margin is removed now in new commit

Rebal67 (approved):

PR #485, opened 2023-02-15
bug/441-private-key-skip  bug removed for updating keys on skip

HaseebUllahAbbasi:


src/containers/FingerprintUpdateContainer.tsx
```diff
@@ -37,8 +37,9 @@ export default function CreateAccountContainer({ password }: { password: string
     };
 
     const onSkip = async () => {
-        await user.saveLocal();
-        updateKeys();
```
theblockstalk, line 41: await updateKeys()

PR #518, opened 2023-02-24
pin screen error

HaseebUllahAbbasi:
Pin Screen Fixed when pressed 3 it returns value of 2 , that is fixed


theblockstalk:
build apk failed. please have a look at this

PR #538, opened 2023-03-02
Feature/234 auth

Rebal67:


src/containers/MainContainer.tsx
```diff
@@ -19,13 +19,29 @@ export default function MainContainer() {
     const [qrOpened, setQrOpened] = useState<boolean>(false);
 
     useEffect(() => {
+        loginToService();
         setUserName();
-        user.communication.onJwtToClient((data) => {
-            console.log(data);
-            navigation.navigate('SSO', { requests: data.requests, platform: 'browser' });
+        user.communication.subscribeMessage((m) => {
```
theblockstalk, line 24: after rebase, this will need to use the new state based navigation

src/containers/MainContainer.tsx
```diff
@@ -19,13 +19,29 @@ export default function MainContainer() {
...
+            navigation.navigate('SSO', {
+                requests: JSON.stringify(message.getPayload().requests),
+                platform: 'browser',
+            });
         });
     }, []);
 
+    //TODO: this should be moved to a store or a provider or a hook
+    async function loginToService() {
+        const message = await user.signMessage({});
```
theblockstalk, line 40: probably want this inside the store, when a user is (attempted) to be initialized from storage. New Task

src/containers/MainContainer.tsx
```diff
@@ -36,8 +52,11 @@ export default function MainContainer() {
         setUsername(u);
     }
 
-    function onScan({ data }: BarCodeScannerResult) {
-        user.communication.connectTonomy(data);
+    async function onScan({ data }: BarCodeScannerResult) {
+        //TODO: change to typed messages
+        const message = await user.signMessage({ type: 'ack' }, data);
+
+        user.communication.sendMessage(message);
```
theblockstalk, line 59: unsure how this function works. maybe some comments?

PR #547, opened 2023-03-06
registration  process

HaseebUllahAbbasi:
Checks if the user has completed all the user OPTIONAL registration steps

### DIFF
--- a/src/containers/CreateAccountUsernameContainer.tsx
+++ b/src/containers/CreateAccountUsernameContainer.tsx
@@ -1,11 +1,11 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 import { TButtonContained } from '../components/atoms/Tbutton';
 import TLink from '../components/atoms/TA';
 import { TCaption, TH1, TP } from '../components/atoms/THeadings';
 import settings from '../settings';
 import { NavigationProp } from '@react-navigation/native';
-import useUserStore from '../store/userStore';
+import useUserStore, { RegLogStatus } from '../store/userStore';
 import { randomString, SdkError, SdkErrors } from '@tonomy/tonomy-id-sdk';
 import TUsername from '../components/TUsername';
 import TInfoBox from '../components/TInfoBox';
@@ -27,13 +27,21 @@ export default function CreateAccountUsernameContainer({ navigation }: Props) {
 
     const errorStore = useErrorStore();
 
-    const { user } = useUserStore();
+    const store = useUserStore();
+
+    useEffect(() => {
+        async function updateStatus() {
+            await store.setRegLogStatus(RegLogStatus.CREATING_ACCOUNT);
+        }
+
+        updateStatus();
+    }, []);
 
     async function onNext() {
         setLoading(true);
 
         try {
-            await user.saveUsername(username);
+            await store.user.saveUsername(username);
         } catch (e: any) {
             if (e instanceof SdkError && e.code === SdkErrors.UsernameTaken) {
                 setErrorMessage('Username already exists');
--- a/src/containers/CreateAccountUsernameContainer.tsx
+++ b/src/containers/CreateAccountUsernameContainer.tsx
@@ -1,11 +1,10 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 import { TButtonContained } from '../components/atoms/Tbutton';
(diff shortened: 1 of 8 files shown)


src/containers/LoginUsernameContainer.tsx
```diff
@@ -24,9 +24,13 @@ export default function LoginUsernameContainer({ navigation }: { navigation: Pro
         },
     });
 
-    const onNext = () => {
-        navigation.navigate('LoginPassword', { username });
-    };
+    useEffect(() => {
+        async function updateStatus() {
```
theblockstalk, line 28: this should only happen when they click on submit their password, that is when they start the PIN flow

src/containers/CreateAccountUsernameContainer.tsx
```diff
@@ -27,13 +27,21 @@ export default function CreateAccountUsernameContainer({ navigation }: Props) {
 
     const errorStore = useErrorStore();
 
-    const { user } = useUserStore();
+    const store = useUserStore();
+
+    useEffect(() => {
+        async function updateStatus() {
+            await store.setRegLogStatus(RegLogStatus.CREATING_ACCOUNT);
```
theblockstalk, line 34: this should only happen when they click on submit their password, that is when they start the PIN flow during create account

src/containers/FingerprintUpdateContainer.tsx
```diff
@@ -51,6 +51,11 @@ export default function CreateAccountContainer({ password }: { password: string
                 await user.saveFingerprint();
                 await user.saveLocal();
                 await updateKeys();
+                await userStore.setBiometric(true);
```
theblockstalk, line 54: we set this, and immediately remove it. So we don't actually need this flag. remove it from the store.

src/store/userStore.ts
```diff
@@ -32,6 +47,35 @@ const userStorage = createStorage<UserStorageState>('tonomyid.user.', storageFac
...
+        await AsynStorage.removeItem('pinStatus');
+        await AsynStorage.removeItem('bioStatus');
+    },
+    getBiometric: async () => {
+        const data = await AsynStorage.getItem('bioStatus');
+
+        if (!data) return null;
+        return JSON.parse(data);
+    },
+    getPin: async () => {
```
theblockstalk, line 66: add typescript that this returns a boolean

src/store/userStore.ts
```diff
@@ -32,6 +47,35 @@ const userStorage = createStorage<UserStorageState>('tonomyid.user.', storageFac
...
+        const data = await AsynStorage.getItem('pinStatus');
+
+        if (!data) return null;
+        return JSON.parse(data);
+    },
+    setPin: async (pinStatus) => {
+        await AsynStorage.setItem('pinStatus', String(pinStatus));
```
theblockstalk, line 73: llets add typescript that pinStatus is a boolean

theblockstalk (changes requested):
This Acceptance criteria is not implemented

- [ ] When the user opens the app THEN Check if the status === creating an account or logging in, and if so take them to the next step they have not finished in those steps
- [ ] if the user has pin flag === false return user to setupPinScreen
- [ ] if the user has biometic flag === false then return to setupBioMetricsScreen

src/store/userStore.ts
```diff
@@ -3,7 +3,14 @@ import RNKeyManager from '../utils/RNKeyManager';
 import { storageFactory } from '../utils/storage';
 import settings from '../settings';
 import { User, createUserObject, setSettings, createStorage } from '@tonomy/tonomy-id-sdk';
+import AsynStorage from '@react-native-async-storage/async-storage';
 
+export enum RegLogStatus {
```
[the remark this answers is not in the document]
HaseebUllahAbbasi: ok I am removing I added because it was in task description

src/containers/LoginUsernameContainer.tsx
HaseebUllahAbbasi: removed and added in password screen

src/containers/CreateAccountUsernameContainer.tsx
HaseebUllahAbbasi: removed and added in password screen

src/containers/FingerprintUpdateContainer.tsx
HaseebUllahAbbasi: removing it from the store

src/store/userStore.ts
HaseebUllahAbbasi: adding typescript

PR #562, opened 2023-03-07
UI change password

StefanoPitto:
Created the navigation to the Change password screen and created the screen to set the new password.


src/containers/SetPasswordContainer.tsx
```diff
@@ -0,0 +1,65 @@
...
+const styles = StyleSheet.create({
+    container: {
+        flex: 1,
+        alignItems: 'center',
+        justifyContent: 'center',
+    },
+    innerContainer: {
+        width: '100%',
+        height: '60%',
+    },
```
Rebal67, line 52: what is the reason of this ?

src/navigation/Settings.tsx
```diff
@@ -0,0 +1,100 @@
...
+        headerTitleStyle: {
+            fontSize: 24,
+            color: 'black',
+        },
+        headerTitleAlign: 'center',
+        headerTintColor: theme.dark ? theme.colors.text : 'black',
+    };
+
+    const noHeaderScreenOptions = { headerShown: false };
+    const CombinedDefaultTheme = merge(navigationTheme, theme);
```
Rebal67, line 70: keep theming at root level if the navigator has theming all child themes will use same theming

src/navigation/Settings.tsx
```diff
@@ -0,0 +1,100 @@
...
+    ConfirmPassword: undefined;
+    SetPassword: undefined;
+};
+
+const Stack = createNativeStackNavigator<RouteStackParamList>();
+
+export default function SettingsNavigation() {
+    const linking = {
+        prefixes: [prefix],
+    };
```
Rebal67, line 29: what does this do ?

src/navigation/Settings.tsx
```diff
@@ -0,0 +1,100 @@
...
+    iosIcon: {
+        paddingBottom: 10,
+        marginLeft: -30,
+        marginTop: -2,
+    },
+    androidIcon: {
+        paddingBottom: 10,
+        marginLeft: -4,
+        marginBottom: -10,
+    },
```
Rebal67, line 99: is negative margin the only way todo this ?

src/containers/SetPasswordContainer.tsx
StefanoPitto: Added this height to the container because it was using 100% of the height so when trying to align it to the center vertically it was in the same position. And if i align the children on this container the inputs losses the width.

src/navigation/Settings.tsx
StefanoPitto: The only way I could make it look as the default stack navigator header.

src/navigation/Settings.tsx
StefanoPitto: removed.

src/navigation/Settings.tsx
StefanoPitto: Removed both lines.

src/navigation/Settings.tsx
Rebal67: but if you didnt change the look of the default wont it choose the default ?

src/containers/SetPasswordContainer.tsx
Rebal67: i dont know if this will result same UI on all devices

src/navigation/Settings.tsx
StefanoPitto: It does not take the default position

src/containers/SetPasswordContainer.tsx
StefanoPitto: Probably on devices with lower resolution it wont make much of a difference and it wont be at the center

PR #563, opened 2023-03-07
Bug/543 create password issue

sadiabbasi:


src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -101,90 +101,101 @@ export default function CreateAccountPasswordContainer({ navigation }: Props) {
 
     return (
         <>
-            <LayoutComponent
-                body={
-                    <View>
-                        <View>
-                            <TH1 style={styles.headline}>Create password</TH1>
+            <ScrollView keyboardShouldPersistTaps={'handled'}>
+                <LayoutComponent
```
Rebal67, line 105: this should be handled globally instead of here.
also no need to add a scroll view to a view where we want to avoid scrolling

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -227,6 +238,9 @@ const errorModalStyles = StyleSheet.create({
 });
 
 const styles = StyleSheet.create({
+    footerText: {
+        marginTop: '30%',
+    },
```
Rebal67, line 243: using margin with percentage will make unpredictable behavior depending on the device size

src/containers/CreateAccountPasswordContainer.tsx
sadiabbasi: I just added the Scroll View to avoid the text overlap in screen.. if i remove Scroll View when user open the keyboard all the screen text move to upward and screen looks not fine https://www.loom.com/share/c8d01b2ee59d465d96ae4632b1fd184b

Rebal67 (approved):

PR #569, opened 2023-03-10
feature/558-change-pin-ui

HaseebUllahAbbasi:
change pin ui


src/navigation/Drawer.tsx
```diff
@@ -40,6 +42,9 @@ export default function DrawerNavigation() {
         >
             {/* change component to Main Navigation when bottom nav should be added */}
             <Drawer.Screen name="UserHome" options={{ title: 'Home' }} component={MainScreen} />
+            {/* <Drawer.Screen name="ChangePin" options={{ title: 'Change Pin' }} component={ChangePinScreen} /> */}
```
theblockstalk, line 45: please remove comments

src/containers/SettingsContainer.tsx
```diff
@@ -23,7 +23,7 @@ export default function SettingsContainer({ navigation }: Props) {
                     />
                     <TNavigationButton
                         onPress={function (): void {
-                            throw new Error('Function not implemented.');
+                            navigation.navigate('ChangePin');
```
theblockstalk, line 26: this is not the design. Design is to go to the next screen
https://www.figma.com/file/cvV48t0f7O2znT6QBxK0Zj/Tonomy-ID?node-id=3336-14957&t=jiCU22LaFu4xgrFQ-4

Its acceptable for now

theblockstalk (changes requested):

src/containers/ChangePinScreenContainer.tsx
```diff
@@ -0,0 +1,144 @@
...
+import useUserStore from '../store/userStore';
+import { HelperText } from 'react-native-paper';
+import LayoutComponent from '../components/layout';
+import useErrorStore from '../store/errorStore';
+import { Props } from '../screens/ChangePinScreen';
+import theme, { commonStyles } from '../utils/theme';
+import settings from '../settings';
+import TInfoBox from '../components/TInfoBox';
+
+export default function ChangePinScreenContainer({ navigation }: { navigation: Props['navigation'] }) {
```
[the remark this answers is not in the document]
HaseebUllahAbbasi: sure making

src/containers/SettingsContainer.tsx
HaseebUllahAbbasi: @theblockstalk 
the 558 and 548 are different task, the #558 is about changing screens where as the #548 is about adding options screen  
the ui you have mentioned above is in #548

src/navigation/Drawer.tsx
```diff
@@ -3,17 +3,15 @@ import { createDrawerNavigator, DrawerNavigationOptions } from '@react-navigatio
...
-import SettingsScreen from '../screens/SettingsScreen';
 import SettingsNavigation from './Settings';
 
 export type RouteDrawerParamList = {
     UserHome: undefined;
     Settings: undefined;
     Help: undefined;
     Logout: undefined;
+    ChangePin: undefined;
+
```
Rebal67, line 14: why change pin is added to the drawer ?

src/navigation/Drawer.tsx
HaseebUllahAbbasi: removed

src/navigation/Drawer.tsx
HaseebUllahAbbasi: it will be merged with 548 later so first 548 is merged with dev then after rebase i will complete its routing,
and these routes will be removed

HaseebUllahAbbasi:
> the task is about making the UI of the PIN UI looks the same. the text isnt the same in 2 pages (when creating account vs login) and there isn't a text when changing a UI plus the buttons aren't like the figma design

the buttons are with border radius and has content and actions according to the conditional rendering , @Rebal67  could you please specify what am i missing , Thanks

Rebal67 (approved):

PR #571, opened 2023-03-13
Feature/445 check key valid

sadiabbasi:


src/containers/MainSplashContainer.tsx
```diff
@@ -2,20 +2,50 @@ import React, { useEffect } from 'react';
...
 import useErrorStore from '../store/errorStore';
 import useUserStore, { UserStatus } from '../store/userStore';
 
 export default function MainSplashScreenContainer({ navigation }: { navigation: NavigationProp<any> }) {
     const errorStore = useErrorStore();
+    const { user, initializeStatusFromStorage, getStatus, setStatus } = useUserStore();
 
-    const { user, initializeStatusFromStorage, getStatus } = useUserStore();
+    const userLogout = () => {
+        user.logout();
```
theblockstalk, line 15: I think all of this should go in the user store, aka a function in `useUserStore().logout()`

src/containers/MainSplashContainer.tsx
```diff
@@ -2,20 +2,50 @@ import React, { useEffect } from 'react';
...
+        setStatus(UserStatus.NOT_LOGGED_IN);
+
+        navigation.dispatch(StackActions.replace('Home'));
```
theblockstalk, line 18: You do not need to dispatch a navigation because by setting UserStatus.NOT_LOGGED_I it will navigate the user to the home screen anyway

src/containers/MainSplashContainer.tsx
```diff
@@ -2,20 +2,50 @@ import React, { useEffect } from 'react';
...
+    };
+    const checkKeys = async () => {
+        await sleep(800);
```
theblockstalk, line 21: why sleep again?

src/containers/MainSplashContainer.tsx
```diff
@@ -2,20 +2,50 @@ import React, { useEffect } from 'react';
...
+
+        try {
+            await user.checkKeysStillValid();
+        } catch (e) {
+            if (e.code === 'KeyNotFound') {
```
theblockstalk, line 26: `e.code === SDKErrors.KeyNotFound` is much better, then we get type safety on the code

src/containers/MainSplashContainer.tsx
```diff
@@ -2,20 +2,50 @@ import React, { useEffect } from 'react';
...
 
     async function main() {
         await sleep(800);
 
+        const username = await user.storage.username;
+        const userStatus = await user.getStatus();
+
+        if (userStatus === 'READY' && UserStatus.LOGGED_IN) {
+            await checkKeys();
+            // if (!getKeyExists()) userLogout();
```
theblockstalk, line 46: remove unused comment. also on L65

src/containers/MainSplashContainer.tsx
sadiabbasi: if i removed this navigation then it's not navigation to home screen it just showing the main splash screen

src/store/userStore.ts
```diff
@@ -50,6 +51,15 @@ const useUserStore = create<UserState>((set, get) => ({
 
         set({ status: status });
     },
+    logout: async () => {
+        set({ status: UserStatus.NOT_LOGGED_IN });
+
+        // Async call to update the status in the storage
+        userStorage.status = UserStatus.NOT_LOGGED_IN;
+        const user = createUserObject(new RNKeyManager(), storageFactory);
```
theblockstalk, line 59: you need to use the user object from the user store and logout of that

theblockstalk: this.user.logout() or something

src/store/userStore.ts
```diff
@@ -50,6 +51,15 @@ const useUserStore = create<UserState>((set, get) => ({
 
         set({ status: status });
     },
+    logout: async () => {
+        set({ status: UserStatus.NOT_LOGGED_IN });
+
+        // Async call to update the status in the storage
+        userStorage.status = UserStatus.NOT_LOGGED_IN;
```
theblockstalk, line 58: better to use `getStatus()` here

theblockstalk: why have L58 and L55. dont they do the same?

src/store/userStore.ts
```diff
@@ -50,6 +51,15 @@ const useUserStore = create<UserState>((set, get) => ({
 
         set({ status: status });
     },
+    logout: async () => {
+        set({ status: UserStatus.NOT_LOGGED_IN });
```
theblockstalk, line 55: better use `setStatus()`

src/store/userStore.ts
sadiabbasi: this.user is not working but i removed the duplication..  getting error i.e. Object is possibly undefined

theblockstalk (approved):

PR #576, opened 2023-03-16
Feature/155 re establish connection

sadiabbasi:


theblockstalk (approved):

src/store/userStore.ts
```diff
@@ -43,6 +43,8 @@ const useUserStore = create<UserState>((set, get) => ({
         userStorage.status = newStatus;
     },
     initializeStatusFromStorage: async () => {
+        await get().user.intializeFromStorage();
```
theblockstalk, line 46: looks like some of this logic is running on other places as well in PR https://github.com/Tonomy-Foundation/Tonomy-ID/pull/571/files#diff-f370f262227d9e004c0126fa7a26c164be8c8bd95a690ae2144e910267fa3f8bR15

^ this is approved, please merge, rebase and then modify this PR so that the same logic will apply but is only executed in on place
- checkKeysStil...() is run inside `user.initializeFromStorage()`
- error handling is moved so that error modal is displayed if checkKeys..() error during iniitialize

PR #577, opened 2023-03-16
feat: ui-for-change-pin-options

HaseebUllahAbbasi:


sadiabbasi (approved):

src/navigation/Pin.tsx
Rebal67, line 1: cant all of them be at the settings navigator ? as the navigation type doesnt change?

Rebal67 (commented):
as stated in the design in Figma there is space between the header and the navigators
also the color of the active buttons are bolder than the disabled buttons can you add that

src/navigation/Pin.tsx
```diff
@@ -0,0 +1,77 @@
...
+import PinSettingsScreen from '../screens/PinSettingsScreen';
+
+export type RouteStackParamList = {
+    AddPin: undefined;
+    Splash: undefined;
+    Settings: undefined;
+    PinSettings: undefined;
+    ChangePin: undefined;
+    RemovePin: undefined;
+};
```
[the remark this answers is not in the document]
HaseebUllahAbbasi: removed

src/navigation/Pin.tsx
HaseebUllahAbbasi: the navigation is nested here so thats why i used other here,

PR #611, opened 2023-03-23
Feature549/UI biometrics implemented

StefanoPitto:


src/containers/SettingsContainer.tsx
```diff
@@ -28,13 +29,23 @@ export default function SettingsContainer({ navigation }: Props) {
...
+                            title={'Fingerprint'}
+                            icon={'fingerprint'}
+                        />
+                    ) : (
+                        <TNavigationButton
+                            onPress={() => {
+                                navigation.navigate('FaceIdSettings');
+                            }}
+                            title={'FaceID'}
+                            icon={<FaceIdIcon style={{ maxWidth: 24, marginHorizontal: 14 }} />}
```
sadiabbasi, line 46: Please avoid inline styling..

src/components/TNavigationButton.tsx
```diff
@@ -19,11 +20,12 @@ export default function TNavigationButton(props: NavigationButtonProps) {
             style={[styles.button, commonStyles.marginBottom]}
             disabled={props.disabled}
         >
-            {props.icon && (
+            {props.icon && typeof props.icon === 'string' && (
```
sadiabbasi, line 23: Why you added typeof here?

sadiabbasi (changes requested):

src/containers/SettingsContainer.tsx
StefanoPitto: removed inline styling

src/components/TNavigationButton.tsx
StefanoPitto: to check if i receive a string or a Component as an icon

sadiabbasi (approved):

PR #630, opened 2023-03-29
fix: :bug: Fixed the app exit on back button

sadiabbasi:


theblockstalk (commented):
does the app still close on the Home screen? We want the app to close on this screen, but on other screens it wont close.

sadiabbasi:
> does the app still close on the Home screen? We want the app to close on this screen, but on other screens it wont close.

yess i tested with rebal.

theblockstalk (approved):

PR #651, opened 2023-04-04
feat: :bug: add check username and error

StefanoPitto:
Now when trying to login and attemping with a wrong username it shows an error.

### DIFF
--- a/src/containers/LoginUsernameContainer.tsx
+++ b/src/containers/LoginUsernameContainer.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { StyleSheet, View, TouchableOpacity } from 'react-native';
 import { useTheme } from 'react-native-paper';
 import LayoutComponent from '../components/layout';
-import { TH1, TP } from '../components/atoms/THeadings';
+import { TCaption, TH1, TP } from '../components/atoms/THeadings';
 import theme, { commonStyles } from '../utils/theme';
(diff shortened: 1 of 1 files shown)


src/containers/LoginUsernameContainer.tsx
```diff
@@ -37,6 +43,7 @@ export default function LoginUsernameContainer({ navigation }: { navigation: Pro
                         <View style={styles.inputContainer}>
                             <TUsername value={username} onChangeText={setUsername} />
                         </View>
+                        <TCaption style={styles.errorStyles}>{errorMessage}</TCaption>
```
Rebal67, line 46: let us make a component or add a mode to caption. 
something like TError or TCaption mode={error}

Rebal67 (approved):

src/containers/LoginUsernameContainer.tsx
```diff
@@ -23,8 +24,13 @@ export default function LoginUsernameContainer({ navigation }: { navigation: Pro
...
     });
 
-    const onNext = () => {
-        navigation.navigate('LoginPassword', { username });
+    useEffect(() => {
+        setErrorMessage('');
+    }, [username]);
+
+    const onNext = async () => {
+        if (await user.usernameExists(username)) navigation.navigate('LoginPassword', { username });
```
theblockstalk, line 32: this should be wrapped in generic error handler

PR #697, opened 2023-05-01
Feature/690 remove pin biometric

sadiabbasi:


src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -107,10 +109,19 @@ export default function CreateAccountPasswordContainer({ navigation }: Props) {
 
     async function onModalPress() {
         setShowModal(false);
-        navigation.navigate('CreateAccountPin', {
-            password,
-            action: 'CREATE_ACCOUNT',
-        });
+
+        try {
+            await user.saveLocal();
```
theblockstalk, line 114: how will the user get feedback that something is happening here? they are in a processing state with no feedback

theblockstalk: can we call saveLocal and updateKeys during `onNext()` and then `userStore.setStatus(UserStatus.LOGGED_IN);` when the user clicks the modal OK button?

src/containers/CreateAccountPasswordContainer.tsx
```diff
@@ -55,6 +57,8 @@ export default function CreateAccountPasswordContainer({ navigation }: Props) {
             await user.savePassword(password, { keyFromPasswordFn: generatePrivateKeyFromPassword });
             const res = await user.createPerson();
 
+            await user.saveLocal();
+            await updateKeys();
```
sadiabbasi, line 61: on catch it's already catched the default error

theblockstalk (approved):

PR #752, opened 2023-05-22
Bug/715 qr button

sadiabbasi:


src/containers/MainContainer.tsx
```diff
@@ -171,7 +173,7 @@ const styles = StyleSheet.create({
         width: '50%',
     },
     marginTop: {
-        marginTop: 28,
+        marginTop: 10,
```
theblockstalk, line 176: looks like you either need to do marginTop: 10 or the <ScrollView>, but not both

with marginTop: 10 but not <ScrollView>
with marginTop: 10 and <ScrollView> looks the same


with <ScrollView> and marginTop: 28

sadiabbasi: marginTop was added before to add space of username from top of screen i found it so much extra space so i reduce space between username and top navigation

PR #753, opened 2023-05-23
Feature/745 notifications

Rebal67:


src/services/CommunicationModule.ts
```diff
@@ -44,13 +46,26 @@ export default function CommunicationModule() {
                     payload: base64UrlPayload,
                     platform: 'browser',
                 });
+                sendLoginNotificationOnBackground(payload.requests[0].getPayload().origin);
```
sadiabbasi, line 49: I guess if origin value is undefined it's throw error may be here you can user "?."

PR #766, opened 2023-06-15
Feature/754 expo deploy production

theblockstalk:


theblockstalk:
- so if `--profile demo` is not added to the command, it will always use `production` profile with the `--auto-submit` argument
- `app.config.ts` Can luckily see environment variables in `process.env`, But when running in EAS, these are not visible and the only variables available are the ones passed to it from the build profile in `eas.json`. An alternative actor this was figured out to save and send environment variables to a .env file and send this to EAS, which then needs `dotenv` to read them in `app.config.ts`
- So we need to make sure we pass the variables in `eas.json` and also explicitly call the profile with `--profile`

theblockstalk:
so it seems like sending the file in a .env file and reading it with `dotenv` is working and the build profiles is not working to give the required environment variable to EAS for `app.config.ts`

theblockstalk:
I was mistaken, the `.env` file was NOT working, and the build profile is now working

theblockstalk:
AFAIKT only `eas.json` and the `"build:prepare"` script are different. But still getting this error in EAS

https://expo.dev/accounts/tonomydev/projects/tonomy-id-demo/builds/d29e1a9f-5611-4e18-80b2-dd3855fa50ed

image

theblockstalk:
should try https://docs.expo.dev/build-reference/troubleshooting/#verify-that-your-project-builds-and-runs-locally

theblockstalk:
adding this back into `test.yaml` and then running `act -s heah..ehahe -j build-apk` allowed to test locally, with same errors

```yaml
    build-apk:
        runs-on: ubuntu-20.04
        env:
          NODE_ENV: staging
          EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
        container: cimg/android:2022.12.1-ndk

        steps:
            - uses: actions/checkout@v3

            - name: Build Tonomy ID .apk file for Android
              run: |
                  curl -fsSL https://deb.nodesource.com/setup_18.x | sudo -E bash -
                  sudo apt-get install -y nodejs
                  node -v
                  npm ci
                  npm run build:prepare
                  npx eas build --clear-cache --profile demo --platform android --local
                  find . -name "*.apk"
                  ```

PR #804, opened 2023-08-07
Feature/802 generate apk

sadiabbasi:


.github/workflows/build-apk.yaml
```diff
@@ -0,0 +1,28 @@
...
+  push:
+    branches:
+      - feature/802-generate-apk
+
+jobs:
+  build-android:
+    runs-on: ubuntu-20.04
+    env:
+      NODE_ENV: demo
+      EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
```
theblockstalk, line 13: remove NODE_ENV it is not used
use EXPO_NODE_ENV and also EXPO_PLATFORM env which are both used for build

theblockstalk: also do a rebase to get fix https://github.com/Tonomy-Foundation/Tonomy-ID/pull/806

.github/workflows/build-apk.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: 🚀 Build and publish tonomy app
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.12.1
+      - run: npm ci
+      - run: npm run build:prepare
+      - run: npx eas build --clear-cache --profile demo --platform android --local
```
theblockstalk, line 26: I don't think we can increase runner size
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

.github/workflows/build-apk.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - name: 🚀 Build and publish tonomy app
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.12.1
+      - run: npm ci
+      - run: npm run build:prepare
+      - run: npx eas build --clear-cache --profile development --platform android --local
+
+      - name: Find APK files
+        run: find . -name "*.apk" > apk_files.txt
```
theblockstalk, line 29: `find . -name "*.apk" -exec mv {} TonomyID.apk \;`
remove L31-37

prepare.app.config.ts
```diff
@@ -76,7 +76,7 @@ const expo: ExpoConfig = {
     ],
     extra: {
         eas: {
-            projectId: '48982c5e-41ff-4858-80ea-8e1b64092f21',
```
theblockstalk, line 79: change this back to previous

sadiabbasi: For the TonomyDev expo account i created development slug and used id from the tonomy-id-dev

theblockstalk (approved):

PR #809, opened 2023-08-09
Bug/792 no login request from other account

theblockstalk:


src/services/CommunicationModule.ts
```diff
@@ -46,8 +48,24 @@ export default function CommunicationModule() {
...
 
     function listenToMessages(): number[] {
-        const loginRequestSubscriber = user.communication.subscribeMessage((message) => {
+        const loginRequestSubscriber = user.communication.subscribeMessage(async (message) => {
             try {
+                const senderDid = message.getSender();
+                const { method, id } = parseDid(senderDid);
+
+                // did:jwk is used for the initial login request so is allowed
+                if (method !== 'jwk' && id !== parseDid(await user.getDid()).id) {
```
sadiabbasi, line 57: don't u think we should show user like what's happening here?

theblockstalk: no not yet. maybe when and if we have a chat feature.

however, Something is happening in the data system that I don't know why it - I don't think a message should be allowed to arrive at the other Tonomy ID user and therefore there is a bug somewhere.

sadiabbasi (approved):

PR #826, opened 2023-08-22
Feature/818 fix mobsf issues

theblockstalk:


theblockstalk:
It looks like we can ignore the first warning:

> medium Application vulnerable to Janus Vulnerability
> CERTIFICATE: Application is signed with v1 signature scheme, making it vulnerable to Janus vulnerability on Android 5.0-8.0, if signed only with v1 signature scheme. Applications running on Android 5.0-7.0 signed with v1, and v2/v3 scheme is also vulnerable.

We ARE producing a v1 AND v2 but NOT v3 schema signature. Tonomy ID is targeting Android SDK minimum 21 maximum 33. See the compatibility here:
https://apilevels.com/

https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/1566#issuecomment-1500755336

theblockstalk:
>  medium App can be installed on a vulnerable Android version
> MANIFEST [minSdk=21] This application can be installed on an older version of android that has multiple unfixed vulnerabilities. Support an Android version > 8, API 26 to receive reasonable security updates.

Updated minAndroid Sdk to 26

theblockstalk:
> medium Application Data can be Backed up
> MANIFEST [android:allowBackup=true] This flag allows anyone to backup your application data via adb. It allows users who have enabled USB debugging to copy application data off of the device.

setting allowBackup to false

theblockstalk:
> medium Broadcast Receiver (com.google.firebase.iid.FirebaseInstanceIdReceiver) is Protected by a permission, but the protection level of the permission should be checked.
> MANIFEST Permission: com.google.android.c2dm.permission.SEND [android:exported=true]
A Broadcast Receiver is found to be shared with other apps on the device therefore leaving it accessible to any other application on the device. It is protected by a permission which is not defined in the analysed application. As a result, the protection level of the permission should be checked where it is defined. If it is set to normal or dangerous, a malicious application can request and obtain the permission and interact with the component. If it is set to signature, only applications signed with the same certificate can obtain the permission.

this is coming from the following code in the  AndroidManifest.xml file
```xml
        <receiver android:name="com.google.firebase.iid.FirebaseInstanceIdReceiver" android:permission="com.google.android.c2dm.permission.SEND" android:exported="true">
            <intent-filter>
                <action android:name="com.google.android.c2dm.intent.RECEIVE" />
            </intent-filter>
        </receiver>
```

http://androidpermissions.com/permission/com.google.android.c2dm.permission.SEND

It is unclear if this permission is normal or dangerous
https://developer.android.com/guide/topics/permissions/overview (see types)

I have tried to fix this by explicitly declaring the `com.google.android.c2dm.permission.SEND` permission

theblockstalk:
>  Broadcast Receiver (androidx.profileinstaller.ProfileInstallReceiver) is Protected by a permission, but the protection level of the permission should be checked.
> MANIFEST Permission: android.permission.DUMP [android:exported=true] A Broadcast Receiver is found to be shared with other apps on the device therefore leaving it accessible to any other application on the device. It is protected by a permission which is not defined in the analysed application. As a result, the protection level of the permission should be checked where it is defined. If it is set to normal or dangerous, a malicious application can request and obtain the permission and interact with the component. If it is set to signature, only applications signed with the same certificate can obtain the permission.

See the previous comment

theblockstalk:
> medium MD5 is a weak hash known to have hash collisions.
> CODE MD5 is a weak hash known to have hash collisions.
https://github.com/MobSF/owasp-mstg/blob/master/Document/0x04g-Testing-Cryptography.md#identifying-insecure-andor-deprecated-cryptographic-algorithms-mstg-crypto-4

I analyse the code and looks like `md5` hashing cryptography is used in the following packages:
- `expo`
  - `expo-asset`
  - `@expo/cli`
    - `json-schema-deref-sync`
    - `@expo/rudder-sdk-node` (also dependency of `eas-cli`)

Given that all of the `md5` library usage is part of the `expo` dependency ( not a dev dependency) there is almost nothing we can do about this...

theblockstalk:
> medium Files may contain hardcoded sensitive information like usernames, passwords, keys etc.
> CODE Files may contain hardcoded sensitive information like usernames, passwords, keys etc.
https://github.com/MobSF/owasp-mstg/blob/master/Document/0x05d-Testing-Data-Storage.md#checking-memory-for-sensitive-data-mstg-storage-10

We are currently using the `@react-native-async-storage/async-storage` Library to store several pieces of data in plaintext:
- https://github.com/Tonomy-Foundation/Tonomy-ID/blob/5b60616bd2b0263953d490486e07d40ff8173a24/src/store/userStore.ts#L19 (Login status)
- https://github.com/Tonomy-Foundation/Tonomy-ID-SDK/blob/cee67fa95ab6c212c54cd5b058b4bc4531c5aaea/src/sdk/controllers/user.ts#L66 (User information including PPI: username)

To solve this security warning we probably need to use an encrypted database for the PPI. I see two strategies:
1. Continue to use `async-storage` and as a wrapper that will encrypt and decrypt all data is goes in and out, using the users password
2. migrate to use the [SQL](https://github.com/MobSF/owasp-mstg/blob/master/Document/0x05d-Testing-Data-Storage.md#sqlite-databases-encrypted) or [TypeORM](https://github.com/typeorm/typeorm) -like databases supported, and use the users password to encrypt the database

My preference would be strategy to, as the cryptography is then handled by the database library, and additionally by using a ORM we can use the same storage package in the BrowserStorage, simplifying our storage libraries

theblockstalk:
> medium App uses SQLite Database and execute raw SQL query. Untrusted user input in raw SQL queries can cause SQL Injection. Also sensitive information should be encrypted and written to the database.
> CODE App uses SQLite Database and execute raw SQL query. Untrusted user input in raw SQL queries can cause SQL Injection. Also sensitive information should be encrypted and written to the database.
https://github.com/MobSF/owasp-mstg/blob/master/Document/0x04h-Testing-Code-Quality.md#injection-flaws-mstg-arch-2-and-mstg-platform-2

Adding validation to all of the user input locations
- KeyManager (key and level)
- UserStorage

This is also done in the SDK here
https://github.com/Tonomy-Foundation/Tonomy-ID-SDK/pull/287

PR #838, opened 2023-08-30
Feature/812 user creates passphrase

sadiabbasi:


src/containers/CreatePassphraseScreenContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { TH1, TP } from '../components/atoms/THeadings';
+import settings from '../settings';
+import theme, { commonStyles } from '../utils/theme';
+import TInfoBox from '../components/TInfoBox';
+import LayoutComponent from '../components/layout';
+import { Props } from '../screens/CreatePassphraseScreen';
+import PassphraseBox from '../components/PassphraseBox';
+
+export default function CreatePassphraseScreenContainer({ navigation }: { navigation: Props['navigation'] }) {
+    const [phraseList, setPhraseList] = useState<string[]>(['barn', 'universe', 'plate', 'star', 'pretty', 'gold']);
```
theblockstalk, line 13: this should be empty by default. but prefilled if not production like we do here
https://github.com/Tonomy-Foundation/Tonomy-ID/blob/development/src/containers/CreateAccountPasswordContainer.tsx#L21

Make sure these 6 world are part of the phrase list

sadiabbasi: these boxes are auto filled when clicked on generate right? it's not like writeable?

theblockstalk (approved):

src/containers/CreatePassphraseScreenContainer.tsx
theblockstalk: correct. the text boxes are not editable

src/containers/CreatePassphraseScreenContainer.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+export default function CreatePassphraseScreenContainer({ navigation }: { navigation: Props['navigation'] }) {
+    const [phraseList, setPhraseList] = useState<string[]>(
+        !settings.isProduction() ? ['barn', 'universe', 'plate', 'star', 'pretty', 'gold'] : ['', '', '', '', '', '']
+    );
+
+    async function onNext() {}
+
+    async function regenerate() {
+        setPhraseList(['gold', 'barn', 'star', 'moon', 'sun', 'tree']);
```
theblockstalk, line 20: this should call `user.generateRandomPassphrase()`

src/containers/CreatePassphraseScreenContainer.tsx
```diff
@@ -0,0 +1,122 @@
...
+import settings from '../settings';
+import theme, { commonStyles } from '../utils/theme';
+import TInfoBox from '../components/TInfoBox';
+import LayoutComponent from '../components/layout';
+import { Props } from '../screens/CreatePassphraseScreen';
+import PassphraseBox from '../components/PassphraseBox';
+
+export default function CreatePassphraseScreenContainer({ navigation }: { navigation: Props['navigation'] }) {
+    const [phraseList, setPhraseList] = useState<string[]>(
+        !settings.isProduction() ? ['barn', 'universe', 'plate', 'star', 'pretty', 'gold'] : ['', '', '', '', '', '']
```
theblockstalk, line 14: when `isProduction` then the empty strings should be generated `user.generateRandomPassphrase()`

PR #843, opened 2023-09-06
Feature/813 confirm passphrase

sadiabbasi:


src/components/AutoComplete.tsx
```diff
@@ -3,23 +3,28 @@ import { View, StyleSheet, Text } from 'react-native';
 import { Menu, TextInput } from 'react-native-paper';
 import theme, { customColors } from '../utils/theme';
 import useUserStore from '../store/userStore';
+import usePassphraseStore from '../store/passphraseStore';
 
 interface AutocompleteProps {
```
theblockstalk, line 8: can you make a TSDoc explaining these values?

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,65 @@
+import create from 'zustand';
+import settings from '../settings';
+
+interface PassphraseStoreState {
+    passphraseList: string[];
+    firstWord: string;
```
theblockstalk, line 6: firstWord, secondWord and thirdWord are redundant. They can be derived easily with passphraseList and randomNumbers. Remove them.

also rename randomNumbers ==> randomWordIndexes

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,65 @@
...
+        const { passphraseList } = get();
+
+        return passphraseList[index] === word;
+    },
+    generateRandomNumbers: () => {
+        const { passphraseList } = get();
+        const randomNumbersList = [] as number[];
+
+        while (randomNumbersList.length < 3) {
+            const randomNumber = Math.floor(Math.random() * 6);
```
theblockstalk, line 44: need to use our cryptographically secure random number generator

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,65 @@
...
+    setFirstWord: (word) => set({ firstWord: word }),
+    setSecondWord: (word) => set({ secondWord: word }),
+    setThirdWord: (word) => set({ thirdWord: word }),
+
+    checkWordAtIndex: (index, word) => {
+        const { passphraseList } = get();
+
+        return passphraseList[index] === word;
+    },
+    generateRandomNumbers: () => {
```
theblockstalk, line 39: maybe rename to `generate3PassphraseIndexes()` would be clearer

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,65 @@
...
+    setFirstWord: (firstWord: string) => void;
+    setSecondWord: (secondWord: string) => void;
+    setThirdWord: (thirdWord: string) => void;
+    checkWordAtIndex: (index: number, word: string) => boolean;
+    generateRandomNumbers: () => void;
+}
+
+type PassphraseStore = PassphraseStoreState & PassphraseStoreActions;
+
+const usePassphraseStore = create<PassphraseStore>((set, get) => ({
```
theblockstalk, line 23: would be a good idea to move the function to generate a new wordlist to be here in the store

src/containers/ConfirmFirstPassphraseWordContainer.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
```
theblockstalk, line 1: all three containers are almost the exact same, except for the index. Combine together and simplify, remember DRY (Don't Repeat Yourself)

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,65 @@
...
+    passphraseList: ['', '', '', '', '', ''],
```
theblockstalk, line 24: you can generate the first wordlist as the default when store is initialized.

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,65 @@
...
+    firstWord: '',
+    secondWord: '',
+    thirdWord: '',
+    randomNumbers: [],
```
theblockstalk, line 28: randomNumbers can also be set when the store is initalized

src/containers/ConfirmFirstPassphraseWordContainer.tsx
theblockstalk: use parameters to control the index and content differences

src/store/passphraseStore.ts
sadiabbasi: when i call get().user.generateRandomPassphrase() it throws error.

src/components/AutoComplete.tsx
```diff
@@ -3,23 +3,36 @@ import { View, StyleSheet, Text } from 'react-native';
...
 import useUserStore from '../store/userStore';
+import usePassphraseStore from '../store/passphraseStore';
 
+/**
+ * Represents an Autocomplete component.
+ * This component provides an input field with autocompletion functionality.
+ * It allows users to input text and provides suggestions based on the input.
+ *
+ * @component
+ * @param {string | undefined} props.value - The default value of the Autocomplete input.
```
theblockstalk, line 14: `@param {string} [value]`

would be better. [] means it's optional. Same for L15 and L16

theblockstalk: what does this value variable mean? "default value" is not informative. Is this the text to search the autocomplete with?

src/components/AutoComplete.tsx
```diff
@@ -3,23 +3,36 @@ import { View, StyleSheet, Text } from 'react-native';
...
+ * @param {number | undefined} props.index - An index to uniquely identify in which confirm word component.
```
theblockstalk, line 15: why does this component need to know about outside components (index)? I think this goes against the Single Responsibility Principal https://www.digitalocean.com/community/conceptual-articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#single-responsibility-principle

This component needs to be used in multiple locations, and does not need to know where it will be implemented.

theblockstalk: most text input (including dropdowns and autoselect) will have an interface:

```ts
interface AutocompleteProps {
  label?: string;
  value?: string;
  onChange?: (value: string) => void
}
```

e.g. https://callstack.github.io/react-native-paper/docs/components/TextInput/#usage

src/containers/ConfirmPassphraseContainer.tsx
```diff
@@ -17,17 +36,28 @@ export default function ConfirmPassphraseContainer() {
                         <View style={{ marginTop: 60 }}>
                             <View style={styles.innerContainer}>
                                 <TP style={styles.textStyle}>
-                                    Please enter the <TP style={styles.boldText}>3rd word</TP> in your passphrase.
+                                    Please enter the{' '}
+                                    <TP style={styles.boldText}>{randomWordIndexes[index] + 1} word</TP> in your
```
theblockstalk, line 40: This should say 1st, 3rd or 5th (with the suffix e.g. "st"). You may need to make a function to create this
1 = st
2 = nd
3 = rd
else  = th

theblockstalk: make a new util for this

src/components/AutoComplete.tsx
theblockstalk: suggest you change setPassphraseValue to onChange

src/containers/ConfirmPassphraseContainer.tsx
```diff
@@ -55,7 +85,7 @@ const styles = StyleSheet.create({
         fontSize: 14,
         fontWeight: '500',
         marginBottom: 8,
-        color: theme.colors.grey1,
+        color: '#5B6261',
```
theblockstalk, line 88: theme color

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,77 @@
...
+import settings from '../settings';
+import { storageFactory } from '../utils/storage';
+import RNKeyManager from '../utils/RNKeyManager';
+import { User, createUserObject } from '@tonomy/tonomy-id-sdk';
+import AsyncStorage from '@react-native-async-storage/async-storage';
+
+interface PassphraseStoreState {
+    user: User;
+    passphraseList: string[];
+    defaultPassphraseWord: string[]; // Store entered words in an array
```
theblockstalk, line 11: this does not need to be stored. you can just use checkWordAtIndex() to check if they remember it and then move on. `passphraseList` remembers all the values.

Also calling this "default" is confusing, the variable doesn't have anything to do with default.

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,77 @@
+import create from 'zustand';
+import settings from '../settings';
+import { storageFactory } from '../utils/storage';
+import RNKeyManager from '../utils/RNKeyManager';
+import { User, createUserObject } from '@tonomy/tonomy-id-sdk';
+import AsyncStorage from '@react-native-async-storage/async-storage';
+
+interface PassphraseStoreState {
+    user: User;
```
theblockstalk, line 9: We should not need the user object to complete this store. Better if just use the util functions from the SDK directly (remove them from the user class - they don't have something to do with the user) and use them directly from the exported util class.

src/store/passphraseStore.ts
theblockstalk: when you remove dependency on the user object this I hope would resolve

theblockstalk:
```ts
if (!settings.isProduction())
  return ['above', 'day', 'fever', 'lemon', 'piano', 'sport']
else
  return get().generateRandomPassphrase()
}
```

This is useful to have a default list for development environment, then we can use the same one for login and be able to test the app faster.

src/components/AutoComplete.tsx
```diff
@@ -3,23 +3,36 @@ import { View, StyleSheet, Text } from 'react-native';
...
 
+/**
+ * Represents an Autocomplete component.
+ * This component provides an input field with autocompletion functionality.
+ * It allows users to input text and provides suggestions based on the input.
+ *
+ * @component
+ * @param {string} [value] - The default value of the Autocomplete input.
+ * @param {string} [onChange] - A function to set the value of the field onChange
+ * @param {number} [screenNumber] - this screenNumber used to enable/disabled the next button so we can check on which screen number user exists then we match word with that index value in passphraseList
```
theblockstalk, line 16: this should not be a Prop. See https://github.com/Tonomy-Foundation/Tonomy-ID/pull/843#discussion_r1317732436

src/components/AutoComplete.tsx
```diff
@@ -81,8 +93,18 @@ const Autocomplete: React.FC<AutocompleteProps> = ({ value: defaultValue, label
...
                                         style={[{ width: '100%' }]}
                                         onPress={() => {
-                                            setValue(word);
                                             setMenuVisible(false);
+                                            onChange(word);
+                                            setErrorMsg('');
+
+                                            if (
+                                                screenNumber &&
+                                                !checkWordAtIndex(randomWordIndexes[screenNumber], word)
```
theblockstalk, line 102: this should be moved out of this component. It is not part of an autosuggest to know what the value should be

src/store/passphraseStore.ts
```diff
@@ -0,0 +1,57 @@
...
+    checkWordAtIndex: (index, word) => {
+        const { passphraseList } = get();
+
+        return passphraseList[index] === word;
+    },
+    generate3PassphraseIndexes: (): number[] => {
+        const randomWordIndexesList = [] as number[];
+
+        while (randomWordIndexesList.length < 3) {
+            const randomValue = randomNumber(0, 6);
```
theblockstalk, line 44: should this not be `randomNumber(0, 5)`? this will have a range of 6

src/utils/passphrase.ts
theblockstalk, line 1: functions here have nothing to do with passphrases. they are about numbers, I would call it numbers.ts

src/utils/passphrase.ts
```diff
@@ -0,0 +1,10 @@
+export function displayScreenNumber(value: number) {
```
theblockstalk, line 1: rename `numberToOrdinal(value: number): string`
https://www.mathsisfun.com/numbers/cardinal-ordinal-chart.html

src/store/passphraseStore.ts
sadiabbasi: i tried this when i call 0,4 it's not generating 5 when i call 6 it's generate betwen 0, 5

PR #846, opened 2023-09-08
Feature/813 confirm passphrase fixed staging crash

sadiabbasi:


theblockstalk (approved):

src/store/passphraseStore.ts
```diff
@@ -19,17 +19,19 @@ interface PassphraseStoreActions {
 type PassphraseStore = PassphraseStoreState & PassphraseStoreActions;
 
 const usePassphraseStore = create<PassphraseStore>((set, get) => ({
-    passphraseList: !settings.isProduction()
-        ? ['above', 'day', 'fever', 'lemon', 'piano', 'sport']
-        : get().generatePassphraseWordList(),
+    passphraseList: ['', '', '', '', '', ''],
```
sadiabbasi, line 22: while initialization when call the get().generatePasshras....() it throws error that it's not defined  ERROR  TypeError: Cannot read property 'generatePassphraseWordList' of undefined, js engine: hermes
 ERROR  Invariant Violation: "main" has not been registered. This can happen if:

PR #857, opened 2023-09-18
Feature/815 login with passphrase

sadiabbasi:


src/store/passphraseStore.ts
```diff
@@ -36,6 +38,7 @@ export const DEFAULT_DEV_PASSPHRASE_LIST = ['above', 'day', 'fever', 'lemon', 'p
 const usePassphraseStore = create<PassphraseStore>((set, get) => ({
     passphraseList: settings.isProduction() ? util.generateRandomKeywords() : DEFAULT_DEV_PASSPHRASE_LIST,
     randomWordIndexes: generate3PassphraseIndexes(),
+    confirmPassphraseWords: ['', '', ''],
```
theblockstalk, line 41: this should be initialized to have the correct words from the `passphraseList`

src/store/passphraseStore.ts
```diff
@@ -6,13 +6,15 @@ import { ApplicationErrors, throwError } from '../utils/errors';
...
+    confirmPassphraseWords: string[];
 }
 
 interface PassphraseStoreActions {
     getPassphrase: () => string;
     generatePassphraseList: () => void;
     unsetPassphraseList: () => void;
     checkWordAtIndex: (index: number, word: string) => boolean;
+    setConfirmPassphraseWord: (index: number, word: string) => void;
 }
```
theblockstalk, line 18: should have an additional function to `unsetConfirPassphraseWord()` which will reset the words back to the default. Otherwise there is a potential security issue as the users root security passphrase is partially exposed - which significantly errodes the security as the other 3 words can be guessed much easier

src/store/passphraseStore.ts
sadiabbasi: but i guess when user go on confirm passphrase screen initially it's empty

theblockstalk: not in developer environment. Then we pre-fill this out. If that is happening already then this is fine. Otherwise we want so that in developer environment it is prefilled and when `isProduction()` true then it is bank

src/store/passphraseStore.ts
```diff
@@ -15,6 +15,7 @@ interface PassphraseStoreActions {
     unsetPassphraseList: () => void;
     checkWordAtIndex: (index: number, word: string) => boolean;
     setConfirmPassphraseWord: (index: number, word: string) => void;
+    unsetConfirmPassphraseWord: () => void;
```
theblockstalk, line 18: you need to call this function at the end of the confirm passphrase flow

src/store/passphraseStore.ts
sadiabbasi: const [value, setValue] = useState<string>(
        settings.isProduction() ? confirmPassphraseWords[index] : passphraseList[randomWordIndexes[index]]
    ); it have like this.

theblockstalk (approved):

PR #876, opened 2023-10-12
Added node version in demo profile

sadiabbasi:


eas.json
```diff
@@ -11,6 +11,7 @@
             "distribution": "internal"
         },
         "demo": {
+            "node": "18.17.1",
```
theblockstalk, line 14: probably want this on all build profiles. keeps everything consistent

theblockstalk (changes requested):

theblockstalk:
sorted this out in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/883/files

PR #884, opened 2023-10-23
Feature/854 profile preview

sadiabbasi:


src/containers/ProfilePreviewContainer.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <>
+            <LayoutComponent
+                body={
+                    <View>
+                        <View style={styles.container}>
+                            <TP size={1} style={styles.textColor}>
+                                Username
+                            </TP>
+                            <View>
+                                <TUsername value={'jack.tanner'} disabled={true} />
```
theblockstalk, line 25: get this from the `userStore.user.getUsername.getBaseUsername()`

src/containers/ProfilePreviewContainer.tsx
```diff
@@ -0,0 +1,92 @@
...
+                        </View>
+                        <View style={[styles.container, styles.marginTop]}>
+                            <TP size={1} style={styles.textColor}>
+                                Anonymous User ID
+                                <TouchableOpacity onPress={() => setInformationPopup(true)}>
+                                    <Text style={styles.informationIcon}>?</Text>
+                                </TouchableOpacity>
+                            </TP>
+                            <View>
+                                <TUsername value={'6845 6863 54C6 D346 '} disabled={true} />
```
theblockstalk, line 36: 1. Should not use TUsername component - it is confusing for devs that don't know this is not a username. Should create a new component at least. It can be just a transfer like `TInputBox = TUsername`
2. `value={await user.getAccountName().toString()}`

src/containers/ProfilePreviewContainer.tsx
```diff
@@ -0,0 +1,92 @@
...
+                visible={informationPopup}
+                iconColor={theme.colors.primary}
+                icon="exclamation"
+                title="Anonymus User ID"
+                onPress={() => setInformationPopup(false)}
+            >
+                <View>
+                    <Text>
+                        An Anonymous User ID conceals your personal info, ensuring your transactions and interactions
+                        remain discreet.
```
theblockstalk, line 68: should be on previous line

sadiabbasi: it doesn't matter if keeps all the space and after utilize full line text go to other line

theblockstalk (approved):

PR #885, opened 2023-10-24
Feature/855 consent screen UI

sadiabbasi:


src/services/CommunicationModule.ts
```diff
@@ -72,10 +72,18 @@ export default function CommunicationModule() {
...
-                navigation.navigate('SSO', {
-                    payload: base64UrlPayload,
-                    platform: 'browser',
-                });
+                if (payload?.requests[0]?.getType() === 'DataSharingRequest') {
+                    navigation.navigate('DataSharingConsent', {
+                        payload: base64UrlPayload,
+                        platform: 'browser',
+                    });
+                } else {
```
theblockstalk, line 80: User will be directed to
- SSO screen if there is a LoginRequest
- DataSharingConsent screen after the SSO screen if there is a DataSharingRequest
- DataSharingConsent scene if there is a DataSharingRequest but not a LoginRequest

sadiabbasi: I removed this from this PR we will manage this in other task as it's UI task

PR #900, opened 2023-11-11
Added loader in Captcha

YAS-SIIN:


src/containers/HcaptchaContainer.tsx
```diff
@@ -170,15 +175,19 @@ export default function HcaptchaContainer({ navigation }: { navigation: Props['n
                                     <View style={{ flexDirection: 'row', alignItems: 'center' }}>
                                         <Checkbox.Android
                                             status={code ? 'checked' : 'unchecked'}
-                                            onPress={() => {
+                                            onPress={() => { 
+                                                setSuccess(!success);
```
theblockstalk, line 179: Better is `setSuccess(success => !success)` because useState hooks are async

theblockstalk (commented):
does this also remove the grey screen? that is the bigger issue

src/containers/HcaptchaContainer.tsx
YAS-SIIN: Done.

YAS-SIIN: let me find other solution about the gray screen

theblockstalk (approved):

PR #951, opened 2024-03-17
Development

sadiabbasi:


src/containers/HcaptchaContainer.tsx
```diff
@@ -50,17 +50,19 @@ export default function HcaptchaContainer({ navigation }: { navigation: Props['n
 
                 setErrorMsg('Challenge expired or some error occured. Please try again.');
             } else {
-                if (settings.config.loggerLevel === 'debug')
+                if (settings.config.loggerLevel === 'debug' || settings.env === 'development') {
+                    setCode('10000000-aaaa-bbbb-cccc-000000000001');
```
theblockstalk, line 54: this does not look correct that the condition statement is based on logger level and env.

logging should depend on logger level, `setCode()` should be based on env.

sadiabbasi: https://github.com/Tonomy-Foundation/Tonomy-ID/pull/952

PR #952, opened 2024-03-18
bug/944-hcaptcha-error

sadiabbasi:


src/containers/HcaptchaContainer.tsx
```diff
@@ -33,30 +33,30 @@ export default function HcaptchaContainer({ navigation }: { navigation: Props['n
...
             } else {
-                if (settings.config.loggerLevel === 'debug' || settings.env === 'development') {
-                    setCode('10000000-aaaa-bbbb-cccc-000000000001');
+                if (settings.config.loggerLevel === 'debug') {
                     console.log('Verified code from hCaptcha', event.nativeEvent.data.substring(0, 10) + '...');
-
-                    if (captchaFormRef.current) {
-                        captchaFormRef.current.hide();
-                    }
+                } else if (settings.env === 'development') {
```
theblockstalk, line 57: what about the old `settings.env === 'local'`? why not use this?

sadiabbasi: when i console it says settings.env equals development that's why i changed it

src/containers/HcaptchaContainer.tsx
```diff
@@ -33,30 +33,30 @@ export default function HcaptchaContainer({ navigation }: { navigation: Props['n
...
-                if (captchaFormRef.current) {
-                    captchaFormRef.current.hide();
-                }
+                hideHcaptcha();
 
                 setErrorMsg('Challenge expired or some error occured. Please try again.');
             } else {
-                if (settings.config.loggerLevel === 'debug' || settings.env === 'development') {
-                    setCode('10000000-aaaa-bbbb-cccc-000000000001');
+                if (settings.config.loggerLevel === 'debug') {
```
theblockstalk, line 55: if (logger level condition)
else if (env condition)


I don't think these should be in the same condition logic flow. Do you?
logging is about logging level. hCaptcha is about environment. Dont mix the two together.

Please check the fix works.

PR #956, opened 2024-04-05
Feature/165 mainnet launch

sadiabbasi:


eas.json
theblockstalk, line 1: why this? seems like we can use pangeaProduction...

.github/workflows/deploy-production.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
       # https://expo.dev/accounts/[account]/settings/access-tokens
       GOOGLE_SERVICE_ACCOUNT_KEY: ${{ secrets.GOOGLE_SERVICE_ACCOUNT_KEY }}
       # https://github.com/expo/fyi/blob/main/creating-google-service-account.md
-      APPLE_STORE_CONNECT_API_KEY: ${{ secrets.APPLE_STORE_CONNECT_API_KEY }}
+      APPLE_STORE_CONNECT_API_KEY_UNITED_WALLET: ${{ secrets.APPLE_STORE_CONNECT_API_KEY_UNITED_WALLET }}
```
theblockstalk, line 16: why?

sadiabbasi: created new app for united wallet that's why

theblockstalk (approved):

PR #967, opened 2024-04-24
added delete popup

sadiabbasi:


src/containers/SettingsContainer.tsx
```diff
@@ -90,11 +97,66 @@ export default function SettingsContainer({ navigation }: { navigation: Props['n
                     title={'Logout'}
                     icon={'logout-variant'}
                 />
+                <TNavigationButton
+                    onPress={() => setShowModal(true)}
+                    title={'Delete Account'}
+                    icon={'delete'}
+                    textColor={theme.colors.error}
```
theblockstalk, line 104: is error the right colour here? the user is not doing something wrong, irreversible and there is not an error

sadiabbasi: error color is set as red.. so in deisgn delete button color is red that's why i used this..

theblockstalk: ok perfect then

PR #985, opened 2024-05-03
Testnet

sadiabbasi:


prepare.app.config.ts
```diff
@@ -14,6 +14,8 @@ console.log('appInputs', appInputs);
 let slug = settings.config.tonomyIdSlug.replace('://', '');
 const scheme = slug;
 
+console.log('scheme', scheme);
```
theblockstalk, line 17: might this not expose sensative information? can we delete in future PR please?

theblockstalk (approved):

PR #1003, opened 2024-05-13
Feature/986 account show

sadiabbasi:


src/utils/numbers.ts
```diff
@@ -8,3 +8,7 @@ export function numberToOrdinal(value: number) {
 
     return screenNumber;
 }
+
+export function formatLEOSAmount(value: number) {
```
theblockstalk, line 12: better call this `formatCurrencyValue()` because you are not just using this for LEOS

PR #1008, opened 2024-05-14
Feature/990 chain abstraction

theblockstalk:
## software API
Polymorphic classes implement chain, account and transaction features needed

## storage
The only thing that needs to be stored to be able to restore an account is the private key.

When the user creates an account or login
1. Generate the deterministic seed with `generateSeedFromPassword()`
2. for each chain
  a. generates the private key from the seed using `generatePrivateKeyFromSeed()`
  b. For each chain, check if the account exists (if needed - Ethereum is not needed)
  c. Store the private key in the persistent storage
  d. create an IAccount for each chain

when the application opens and the App initializes, it should check for stored private keys and create a new IAccount for each chain with an account


sadiabbasi:
> ## software API
> Polymorphic classes implement chain, account and transaction features needed
> 
> ## storage
> The only thing that needs to be stored to be able to restore an account is the private key.
> 
> When the user creates an account or login
> 
> 1. Generate the deterministic seed with `generateSeedFromPassword()`
> 2. for each chain
>    a. generates the private key from the seed using `generatePrivateKeyFromSeed()`
>    b. For each chain, check if the account exists (if needed - Ethereum is not needed)
>    c. Store the private key in the persistent storage
>    d. create an IAccount for each chain
> 
> when the application opens and the App initializes, it should check for stored private keys and create a new IAccount for each chain with an account

is it okay to store private key in the storage?

src/utils/chain/etherum.ts
```diff
@@ -0,0 +1,292 @@
...
+    IToken,
+    TransactionType,
+    AbstractChain,
+    AbstractToken,
+    IAccount,
+    AbstractAccount,
+    ITransaction,
+} from './types';
+
+const ETHERSCAN_API_KEY = 'your-etherscan-api-key';
```
sadiabbasi, line 26: how to get these keys?

src/utils/chain/types.ts
```diff
@@ -0,0 +1,248 @@
...
+    getNativeToken(): IToken {
+        return this.nativeToken;
+    }
+    abstract getTokens(): Promise<IToken[]>;
+    getBalance(token: IToken): Promise<IAsset> {
+        return token.getBalance(this);
+    }
+    abstract signTransaction(transaction: unknown): Promise<unknown>;
+    abstract sendSignedTransaction(signedTransaction: unknown): Promise<unknown>;
+    abstract sendTransaction(transaction: unknown): Promise<unknown>;
```
sadiabbasi, line 247: here why the types are unknown..

src/utils/keys.ts
```diff
@@ -29,22 +30,34 @@ export async function testKeyGenerator() {
...
+    const privateKey = new PrivateKey(KeyType.K1, bytes);
+
+    return {
+        privateKey,
+        salt: Checksum256.from(saltString),
+    };
+}
+
+async function generateSeedFromPassword(password: string, salt?: string): Promise<{ seed: string; salt: string }> {
+    if (!salt) salt = Checksum256.from(randomBytes(32)).hexString;
```
sadiabbasi, line 47: here when we call this in create account generate seed from password without salt it generate random  bytes for salt.
when we call i this in login it generate another randombytes for salt due to this seed result not match..
or will it create the same seed even if the salt value differe?

theblockstalk:
> > ## software API
> > Polymorphic classes implement chain, account and transaction features needed
> > ## storage
> > The only thing that needs to be stored to be able to restore an account is the private key.
> > When the user creates an account or login
> > 
> > 1. Generate the deterministic seed with `generateSeedFromPassword()`
> > 2. for each chain
> >    a. generates the private key from the seed using `generatePrivateKeyFromSeed()`
> >    b. For each chain, check if the account exists (if needed - Ethereum is not needed)
> >    c. Store the private key in the persistent storage
> >    d. create an IAccount for each chain
> > 
> > when the application opens and the App initializes, it should check for stored private keys and create a new IAccount for each chain with an account
> 
> is it okay to store private key in the storage?

depends how the are stored of course. If they are on the users device they are pretty good.
- Especially if there is no API to export them
- especially if they are stored using encryption at rest
- especially as they are stored in protected storage (offered by most mobile operating systems)

src/utils/chain/etherum.ts
theblockstalk: We will need to make an account with etherscan, then import them as environment variables

src/utils/chain/types.ts
theblockstalk: Because each chain, the arguments and the response for signing a transaction are different

src/utils/keys.ts
theblockstalk: To create the same seed it needs the same password and salt.
When we first create the Pangea account, it generates salt that we put on the blockchain. After this (during login) we fetched the salt from the blockchain and use it to generate the same seed every time

src/utils/chain/types.ts
```diff
@@ -0,0 +1,251 @@
...
+    getType(): Promise<TKeyType>;
+    toString(format?: KeyFormat): Promise<string>;
+}
+
+export abstract class AbstractPublicKey implements IPublicKey {
+    protected key: IKey;
+
+    constructor(key: IKey) {
+        this.key = { ...key };
+        delete this.key.privateKeyHex;
```
sadiabbasi, line 16: after store in storage then delete?

src/utils/chain/types.ts
```diff
@@ -0,0 +1,251 @@
...
+    getPublicKey(): Promise<IPublicKey>;
+    signTransaction(transaction: unknown): Promise<unknown>;
+}
+
+export abstract class AbstractPrivateKey implements IPrivateKey {
+    protected abstract kid: string;
+    protected abstract keyManager: IKeyManager;
+
+    getKid(): string {
+        return this.kid;
```
sadiabbasi, line 43: is this return private key?

src/utils/chain/types.ts
```diff
@@ -0,0 +1,251 @@
...
+        return this.token.getPrecision();
+    }
+    printValue(): string {
+        const value = this.amount / BigInt(10 ** this.token.getPrecision());
+
+        return value.toString();
+    }
+    toString(): string {
+        return `${this.printValue()} ${this.token.getSymbol()}`;
+    }
```
sadiabbasi, line 136: here may be we can add gas price calculation function?

theblockstalk:
https://github.com/Tonomy-Foundation/Tonomy-ID-SDK/issues/336 this issue is blocking the Veramo library from compiling. So I reverted to not use these libraries

PR #1012, opened 2024-05-22
added key storage in typeorm

sadiabbasi:


babel.config.js
```diff
@@ -6,6 +6,10 @@ module.exports = function (api) {
             'react-native-paper/babel',
             'transform-inline-environment-variables',
             'react-native-reanimated/plugin',
+            '@babel/transform-react-jsx-source',
```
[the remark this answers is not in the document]
sadiabbasi: yes it throws error when i install typeorm package then i added these..

src/containers/HcaptchaContainer.tsx
```diff
@@ -91,6 +93,9 @@ export default function HcaptchaContainer({ navigation }: { navigation: Props['n
         }
 
         try {
+            const ethereumChain = new EthereumChain();
+
+            savePrivateKeyToStorage(getPassphrase(), ethereumChain);
```
theblockstalk, line 98: should be EthereumChain for production, EthereumSepoliaChain for staging/testnet

src/utils/chain/etherum.ts
```diff
@@ -80,7 +80,12 @@ export class EthereumChain extends AbstractChain {
     protected name = 'Ethereum';
     protected chainId = '1';
     protected logoUrl = 'https://cryptologos.cc/logos/ethereum-eth-logo.png';
-    protected nativeToken = new ETHToken();
+    protected nativeToken: ETHToken;
```
theblockstalk, line 83: `protected nativeToken = new ETHToken();`
is a bit cleaner than having it in the constructor

src/utils/chain/etherum.ts
```diff
@@ -105,9 +110,14 @@ export class ETHToken extends AbstractToken {
     protected name = 'Ether';
     protected symbol = 'ETH';
     protected precision = 18;
-    protected chain = new EthereumChain();
+    protected chain: EthereumChain;
     protected logoUrl = 'https://cryptologos.cc/logos/ethereum-eth-logo.png';
 
+    constructor(chain: EthereumChain) {
```
theblockstalk, line 116: wondering if EthereumChain, ETHToken should be singletons, not dynamic classes. Would simplify things quite a bit...

Or alternative we make a dynamic Chain and Token class, which we can then quickly create new chains with using `new` operator

src/veramo/setup.ts
```diff
@@ -0,0 +1,16 @@
...
+export const dataSource = new DataSource({
+    database: 'storage',
+    driver: ExpoSQLiteDriver,
+    entities: [keyStorage],
+    synchronize: false,
+    type: 'expo',
+});
+
+export function connect() {
+    if (dataSource.isInitialized) return Promise.resolve(dataSource);
```
theblockstalk, line 14: `return await dataSource` would be cleaner, and force the function to be `async` which gives type safety

src/veramo/entities/keyStorage.ts
theblockstalk, line 1: directory name still "veramo"

src/veramo/repositories/storageRepository.ts
theblockstalk, line 1: I think we want another abstraction layer on top of this, which then uses our new Key/Transaction/Chain abstraction types. aka

class KeyManager {
getAll(): Promise<PublicKey>
create(name: string: privateKey: PrivateKey): Promise<void>
...
}

src/utils/keys.ts
```diff
@@ -66,9 +70,39 @@ async function generateSeedFromPassword(password: string, salt?: string): Promis
     return { seed: result.rawHash as string, salt };
 }
 
-// async function generatePrivateKeyFromSeed(seed: string, chain: IChain): Promise<IPrivateKey> {
-//     const chainSeed = sha256(seed + chain.getChainId());
+async function generatePrivateKeyFromSeed(
+    seed: string,
+    chain: IChain
+): Promise<{ privateKeyHex: string; privateKey: IPrivateKey; ethereumAccount: string }> {
```
theblockstalk, line 76: this should not be ethereum specific. that is the whole point of the abstractions we built.

the user should be able to call `chain.createKeyFromSeed(seed)` which will work on any chain we send in

```ts
   const privateKey = chain.createKeyFromSeed(chainSeed)

    return {
        privateKey,
    };
```

src/utils/keys.ts
```diff
@@ -66,9 +70,39 @@ async function generateSeedFromPassword(password: string, salt?: string): Promis
...
+    const ethereumAccount = new EthereumAccount(await privateKey.getAddress(), privateKey).getName();
+
+    return {
+        privateKeyHex,
+        privateKey,
+        ethereumAccount,
+    };
+}
+
+export async function savePrivateKeyToStorage(passphrase: string, chain: IChain, salt?: string): Promise<void> {
```
theblockstalk, line 97:
```ts
export async function savePrivateKeyToStorage(passphrase: string, salt?: string): Promise<void> {
    const dataSource = await connect(); // THIS SHOULD BE DONE ON APP INITIALIZATION, NOT HERE
    const keyStorageRepo = new keyStorageRepository(dataSource);

    const seedData = await generateSeedFromPassword(passphrase, salt);

    const ethereumKey = await generatePrivateKeyFromSeed(passphrase, new EthereumChain());
    const hederaKey = await generatePrivateKeyFromSeed(passphrase, new HederaChain()));

    // Save the key and seed to the keyStorage
    await keyStorageRepo.create('seed', seedData.seed);
    await keyStorageRepo.create('ethereum', ethereumKey); // we should use the new abstraction layer I mention
    await keyStorageRepo.create('hedera', hederaKey);
}

File: src/veramo/repositories/storageRepository.ts
Comment (sadiabbasi) [reply src/veramo/repositories/storageRepository.ts:1]: in create i added the type to PrivateKey but when trying to store/create seed it throws error

File: src/utils/chain/etherum.ts
Comment (theblockstalk) [inline src/utils/chain/etherum.ts:1]: see https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1012/commits/58136746fb0d8c0bd53c196590b3cf3ea53633cb

Hope makes sense. Was too hard to add as comments here.

File: src/utils/chain/etherum.ts [hunk 8]
```diff
@@ -22,7 +23,8 @@ import {
     AbstractAsset,
     IPrivateKey,
 } from './types';
-import { TKeyType } from '@veramo/core';
+import { Bytes, KeyType, PrivateKey } from '@wharfkit/antelope';
```
theblockstalk, line 26: should not use this library here

src/utils/chain/etherum.ts
```diff
@@ -76,14 +78,34 @@ export class EthereumPrivateKey extends AbstractPrivateKey implements IPrivateKe
...
+        const bytes = Bytes.from(chainSeed, 'hex');
+        const privateKeyValue = new PrivateKey(KeyType.K1, bytes);
+        const privateKeyHex = '0x' + sha256(privateKeyValue.toString());
+
+        return new EthereumPrivateKey(privateKeyHex);
+    }
 }
 
-export class ETHAsset extends AbstractAsset {
+export class Asset extends AbstractAsset {
```
theblockstalk, line 108: should move Asset and Token to the types.ts file, as they are now not Ethereum specific

src/utils/chain/etherum.ts
theblockstalk: Consumer files will need to consume these differently.

e.g. they should import EthereumMainnetChain and ETHToken

src/utils/chain/etherum.ts
```diff
@@ -134,20 +167,46 @@ export class ETHToken extends AbstractToken {
     }
 }
 
+const EthereumMainnetChain = new EthereumChain('Ethereum', '1', 'https://cryptologos.cc/logos/ethereum-eth-logo.png');
```
theblockstalk, line 170: now we can dynamically create Ethereum chains easily

src/utils/keys.ts
```diff
@@ -66,9 +71,33 @@ async function generateSeedFromPassword(password: string, salt?: string): Promis
     return { seed: result.rawHash as string, salt };
 }
 
-// async function generatePrivateKeyFromSeed(seed: string, chain: IChain): Promise<IPrivateKey> {
-//     const chainSeed = sha256(seed + chain.getChainId());
+async function generatePrivateKeyFromSeed(seed: string, chain: IChain): Promise<IPrivateKey> {
+    const chainSeed = sha256(seed + chain.getChainId());
+
+    const privateKey = chain.createKeyFromSeed(chainSeed);
```
theblockstalk, line 77: 😄

src/utils/keys.ts
```diff
@@ -66,9 +71,33 @@ async function generateSeedFromPassword(password: string, salt?: string): Promis
...
+    const seedData = await generateSeedFromPassword(passphrase, salt);
+    let ethereumKey;
 
-//     // create privateKey from chainSeed
-//     // return privateKey
-// }
+    if (settings.env === 'staging' || settings.env === 'testnet') {
+        ethereumKey = await generatePrivateKeyFromSeed(
+            passphrase,
+            new EthereumChain('Sepolia', '11155111', 'https://cryptologos.cc/logos/ethereum-eth-logo.png')
```
theblockstalk, line 91: mainnet and sepolia chain now an exported object from ethereum.ts

src/utils/chain/etherum.ts
```diff
@@ -22,7 +23,8 @@ import {
     AbstractAsset,
     IPrivateKey,
 } from './types';
-import { TKeyType } from '@veramo/core';
+import { Bytes, KeyType, PrivateKey } from '@wharfkit/antelope';
+import { sha256 } from '@tonomy/tonomy-id-sdk';
```
theblockstalk, line 27: `const INFURA_URL = `https://mainnet.infura.io/v3/${INFURA_KEY}`;`

infura url might be different per chain, should be part of EthereumChain constructor

src/veramo/repositories/storageRepository.ts
```diff
@@ -0,0 +1,32 @@
+import { Repository, DataSource } from 'typeorm/browser';
+import { keyStorage } from '../entities/keyStorage';
+import { KeyManager } from './keyManager';
+import { IPrivateKey } from '../../utils/chain/types';
+
+export class keyStorageRepository extends KeyManager {
```
theblockstalk, line 6: does not need to extend. this is a layer underneath the key manager

src/veramo/repositories/keyManager.ts
```diff
@@ -0,0 +1,11 @@
+import { keyStorageRepository } from './storageRepository';
+import { keyStorage } from '../entities/keyStorage';
+
+export abstract class KeyManager {
+    protected abstract repository: keyStorageRepository;
+
+    abstract getAll(): Promise<keyStorage[]>;
```
theblockstalk, line 7:
```ts
getPublicKeys(): Promise<IPrivateKey[]>;
addKey(name: string, privateKey: IPrivateKey): Promise<void>;
findByName(name: string): Promise<IPrivateKey>;
delete(name: string): Promise<void>;
```

src/veramo/repositories/keyManager.ts
```diff
@@ -0,0 +1,11 @@
+import { keyStorageRepository } from './storageRepository';
+import { keyStorage } from '../entities/keyStorage';
+
+export abstract class KeyManager {
```
theblockstalk, line 4: this is a layer on top of (not a super class, but a abstraction above) that adds types to the KeyStorageRepository

src/veramo/setup.ts
theblockstalk, line 1: directory name

src/utils/chain/etherum.ts
sadiabbasi: I moved Asset to types but i leave the Token as it used getPrice function that returns ethereum.usd and also use provider.

src/utils/keys.ts
```diff
@@ -66,9 +75,27 @@ async function generateSeedFromPassword(password: string, salt?: string): Promis
...
+
+    const seedData = await generateSeedFromPassword(passphrase, salt);
+    let ethereumKey;
 
-//     // create privateKey from chainSeed
-//     // return privateKey
-// }
+    if (settings.env === 'staging' || settings.env === 'testnet') {
+        ethereumKey = await generatePrivateKeyFromSeed(passphrase, EthereumSepoliaChain);
+    } else {
```
theblockstalk, line 94: safer would be to use mainnet for `settings.env` === production, and everything else Sepolia. That would mean nobody can accidentally use mainnet except on the production env.

src/keyManager/repositories/keyManager.ts
```diff
@@ -0,0 +1,13 @@
+import { KeyStorageRepository } from './storageRepository';
+import { IPrivateKey } from '../../utils/chain/types';
+
+export abstract class KeyManager {
```
theblockstalk, line 4: this should be implemented and used in `savePrivateKeyToStorage()`

src/utils/chain/etherum.ts
theblockstalk: we also need a way to set the infura key

src/keyManager/entities/keyStorage.ts
```diff
@@ -0,0 +1,13 @@
...
+@Entity('keyStorage')
+export class keyStorage {
+    @PrimaryGeneratedColumn()
+    id!: number;
+
+    @Column()
+    name!: string;
+
+    @Column()
+    value!: string;
```
theblockstalk, line 12: could also add

```ts
addedAt: Date
    updatedAt: Date
```

which could be good for tracking later...

src/utils/StorageManager/entities/appStorage.ts
theblockstalk, line 1: This name might be misleading, because we have the app (United Citizens Wallet) and also Apps which Pangea Citizens can launch. Can you make this more specific? AppSettings, or just Settings

src/utils/StorageManager/repositories/KeyStorageRepository.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        return storage;
+    }
+
+    public async findByName(name: string): Promise<KeyStorage | null> {
+        return this.ormRepository.findOne({ where: { name } });
+    }
+
+    public async delete(): Promise<void> {
+        await this.ormRepository.delete({});
```
theblockstalk, line 30: this will delete all right? then it should be called `deleteAll()` or otherwise it should be `deleteByName(name)` or `deleteById(id)`

src/utils/StorageManager/repositories/KeyStorageRepository.ts
```diff
@@ -0,0 +1,32 @@
...
+import { IPrivateKey } from '../../chain/types';
+
+export class KeyStorageRepository {
+    private ormRepository: Repository<KeyStorage>;
+
+    constructor(dataSource: DataSource) {
+        this.ormRepository = dataSource.getRepository(KeyStorage);
+    }
+
+    public async storeKey(name: string, value: string): Promise<KeyStorage> {
```
theblockstalk, line 12: This will always create a new table row. Should use a better function name to indicate this like `storeNewKey()`.

I think we want it that the key `name` is always unique. That way there is no confusion when `findByName()`. Therefore I suggest one of the two strategies:
1. `addKey()` which will throw an if the name already exists + `updateKey()` Which throw if the name does not exist, or
2. `emplaceKey()` went well at a new row or update a key if the name already exists

src/utils/StorageManager/entities/appStorage.ts
```diff
@@ -0,0 +1,19 @@
+import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm/browser';
+
+@Entity('AppStorage')
+export class AppStorage {
+    @PrimaryGeneratedColumn()
+    id!: number;
+
+    @Column()
+    name!: string;
```
theblockstalk, line 9: is there a describtor you can add to force that this is unique here?

src/utils/StorageManager/repositories/appSettingRepository.ts
```diff
@@ -0,0 +1,31 @@
...
+import { AppStorage } from '../entities/appStorage';
+
+export class AppStorageRepository {
+    private ormRepository: Repository<AppStorage>;
+
+    constructor(dataSource: DataSource) {
+        this.ormRepository = dataSource.getRepository(AppStorage);
+    }
+
+    public async storeSetting(name: string, value: string): Promise<AppStorage> {
```
theblockstalk, line 11: similar comments to `src/utils/StorageManager/repositories/KeyStorageRepository.ts`

src/utils/StorageManager/repositories/appStorageManager.ts
```diff
@@ -0,0 +1,24 @@
+import { AppStorageRepository } from './appSettingRepository';
+
+export abstract class AppStorageManager {
+    protected repository: AppStorageRepository;
+
+    constructor(repository: AppStorageRepository) {
+        this.repository = repository;
+    }
+
+    public async addNewSetting(name: string, value: string): Promise<void> {
```
theblockstalk, line 10: At this level you want to use proper types. Let's create specific functions for the specific settings need to be stored

```ts
setCryptoSeed(seed: string): Promise<void> {
   await this.repository.emplaceSeed('seed', seed);
}
getCryptoSeed(): Promise<string | null> {
  const seed = await this.repository.findByName("seed");
  if (seed) return seed.value;
        else return null;
}

File: src/utils/chain/types.ts [hunk 25]
```diff
@@ -45,6 +46,9 @@ export abstract class AbstractPrivateKey implements IPrivateKey {
 
     abstract getPublicKey(): Promise<IPublicKey>;
     abstract signTransaction(transaction: unknown): Promise<unknown>;
+    async toString(): Promise<string> {
```
theblockstalk, line 49: better rename `exportPrivateKey()` to be clear what this gets (otherwise it could be interpreted to be the public key string

src/utils/StorageManager/repositories/keyStorageManager.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public async addKey(name: string, privateKey: IPrivateKey): Promise<void> {
+        await this.repository.storeKey(name, await privateKey.toString());
+    }
+
+    public async findByName(name: string): Promise<IPrivateKey | null> {
+        const key = await this.repository.findByName(name);
+        let privateKey;
+
+        if (key && key.name === 'ethereum') {
```
theblockstalk, line 20: simpler. if a key is found but cannot be made into a IPrivateKey then no point returning it at this abstraction level - it breaks the type expected of


```ts
if (key && key.name === 'ethereum') {
            return ew EthereumPrivateKey(key.value);
        }
else return null
```

src/utils/chain/etherum.ts
```diff
@@ -19,17 +21,15 @@ import {
...
     AbstractPublicKey,
     AbstractPrivateKey,
-    AbstractAsset,
     IPrivateKey,
+    Asset,
 } from './types';
-import { TKeyType } from '@veramo/core';
+import settings from '../../settings';
 
 const ETHERSCAN_API_KEY = 'your-etherscan-api-key';
```
theblockstalk, line 29: should also be a setting

src/utils/keys.ts
```diff
@@ -30,7 +33,12 @@ export async function testKeyGenerator() {
         const privateKeyEth = new EthereumPrivateKey(
             Wallet.fromPhrase('save west spatial goose rotate glass any phrase manual pause category flight').privateKey
         );
-        const ethereumAccount = new EthereumAccount(await privateKeyEth.getAddress(), privateKeyEth);
+
+        const ethereumAccount = new EthereumAccount(
```
theblockstalk, line 37: better use the constructor overload `const ethereumAccount = EthereumAccount.fromPrivateKey(...`

src/utils/StorageManager/entities/appStorage.ts
theblockstalk: also could be good if you can add a descriptor that makes this column indexed for faster lookups

src/utils/StorageManager/repositories/appStorageManager.ts
theblockstalk: for each other setting there would be a getter/setter as well with types

src/utils/StorageManager/repositories/KeyStorageRepository.ts
sadiabbasi: if we delete key from storage on logout.. then their is no key name already exists as we are calling the store only on create account and login..

src/utils/chain/types.ts
```diff
@@ -53,13 +57,15 @@ export interface IChain {
     getLogoUrl(): string;
     // getApiEndpoint(): string;
     getNativeToken(): IToken;
+    createKeyFromSeed(seed: string): IPrivateKey;
+    getInfuraUrl(): string;
```
theblockstalk, line 61: this should not be part of IChain because it is Ethereum specific, it can be in the EthereumChain implementation though

src/utils/chain/types.ts
```diff
@@ -53,13 +57,15 @@ export interface IChain {
...
 }
 
 export abstract class AbstractChain {
```
theblockstalk, line 64: `AbstractChain implements IChain`

src/utils/StorageManager/repositories/keyStorageManager.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public async emplaceKey(name: string, privateKey: IPrivateKey): Promise<void> {
+        const existingKey = await this.repository.findByName(name);
+        const value = await privateKey.exportPrivateKey();
+
+        if (existingKey) {
+            existingKey.value = value;
+            existingKey.updatedAt = new Date();
+            await this.repository.updateKey(existingKey);
```
sadiabbasi, line 19: for updateKey it's not possible to update the id, as we get the existing key and just update the value and updatedAt field here..

theblockstalk (approved):

PR #1016, opened 2024-05-28
Feature/upcoming session

sadiabbasi:


src/services/WalletConnect/WalletConnectModule.ts
```diff
@@ -0,0 +1,54 @@
...
+    let ethereumAccount;
+
+    if (privateKey) {
+        const ethereumPrivateKey = new EthereumPrivateKey(privateKey?.privateKeyHex); // Cast privateKey to EthereumPrivateKey
+
+        ethereumAccount = await EthereumAccount.fromPublicKey(
+            EthereumSepoliaChain,
+            await ethereumPrivateKey.getPublicKey()
+        );
+        currentETHAddress = ethereumAccount.name;
```
theblockstalk, line 31: better use getName() which keeps it consistent with the IAccount interface

src/services/WalletConnect/WalletConnectModule.ts
```diff
@@ -0,0 +1,54 @@
...
+                url: 'https://walletconnect.com/',
+                icons: [settings.config.images.logo48],
+            },
+        });
+    } else {
+        throw new Error('No private key found');
+    }
+}
+
+export async function _pair(params: { uri: string }) {
```
theblockstalk, line 46: why not just have the uri: string as the input?

src/services/CommunicationModule.ts
```diff
@@ -137,5 +139,50 @@ export default function CommunicationModule() {
...
+            case 'eth_sendTransaction':
+                navigation.navigate('SignTransaction', {
+                    requestSession: requestSessionData,
+                    requestEvent: requestEvent,
+                });
+                sendWalletConnectNotificationOnBackground(
+                    'Transaction Request',
+                    'Ethereum transaction signing request'
+                );
+                return;
```
theblockstalk, line 177: switch for only 1 case? should at least have a default case which it would then throw an error or respond to client that it is not available...???

package.json
```diff
@@ -25,13 +25,18 @@
...
         "@react-navigation/native": "^6.0.12",
         "@react-navigation/native-stack": "^6.8.0",
         "@tonomy/tonomy-id-sdk": "0.21.0-development.1",
         "@veramo/core": "^6.0.0",
+        "@walletconnect/core": "^2.13.0",
+        "@walletconnect/react-native-compat": "^2.13.0",
+        "@walletconnect/utils": "^2.13.0",
+        "@walletconnect/web3wallet": "^1.12.0",
         "@wharfkit/antelope": "^0.7.1",
         "babel-plugin-transform-typescript-metadata": "^0.3.2",
```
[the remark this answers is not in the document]
sadiabbasi: added this in babel plugins when working on storage get some error(didn't remember exact error) and google suggested this solution so i added this..

src/containers/WalletConnectLoginContainer.tsx
```diff
@@ -0,0 +1,166 @@
...
+            id: id,
+            reason: getSdkError('USER_REJECTED'),
+        });
+        navigation.navigate('Drawer', { screen: 'UserHome' });
+    };
+
+    async function handleAccept() {
+        const { requiredNamespaces, relays } = params;
+
+        if (payload) {
```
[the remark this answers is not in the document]
sadiabbasi: it's not doing transaction it's creating session.. do you think we still need to add checks?

package.json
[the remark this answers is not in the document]
theblockstalk: yes but bable is used for compilation, thus should be a devDependancy. It is not used at runtime (dependency)

src/containers/WalletConnectLoginContainer.tsx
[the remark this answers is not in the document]
theblockstalk: you are correct. for now its OK. it might also be a good idea then to create a IChainSession interface with derived classes which we can use for this, similar to the ITransaction interface will be used for the transaction signing screens

src/utils/chain/types.ts
```diff
@@ -251,3 +252,12 @@ export class Asset extends AbstractAsset {
         this.amount = amount;
     }
 }
+
+export interface IChainSession {
+    getId(): number;
```
theblockstalk, line 257: this might not always be a number. string is safer

src/utils/chain/types.ts
```diff
@@ -251,3 +252,12 @@ export class Asset extends AbstractAsset {
...
+    getName(): string;
+    getUrl(): string;
+    getIcons(): string[];
```
theblockstalk, line 260: multiple icons? what for?

src/utils/chain/types.ts
```diff
@@ -251,3 +252,12 @@ export class Asset extends AbstractAsset {
...
+    getNamespaces(): SessionTypes.Namespaces;
```
theblockstalk, line 261: we will not use WalletConnect for all Sessions, so cannot be return type. what is `getNamespaces()` used for?

src/utils/chain/types.ts
```diff
@@ -251,3 +252,12 @@ export class Asset extends AbstractAsset {
...
+    verifySession(): Promise<void>;
```
theblockstalk, line 262: maybe instead have two function acceptSession() and rejectSession() ?

src/utils/chain/types.ts
sadiabbasi: used to get namespaces.. for the session proposal it takes accunt name, which method supported etc..

src/utils/chain/types.ts
sadiabbasi: this.payload.params.proposer.metadata.icons this return icons in array

src/utils/chain/types.ts
sadiabbasi: if i change here we need to do type casting as payload.id return number type and here expected number here
 await web3wallet?.approveSession({
            id: this.getId(),
            relayProtocol: this.payload.params.relays[0].protocol,
            namespaces,
        });

PR #1019, opened 2024-05-30
Feature/1013 ethereum storage

sadiabbasi:


src/containers/CreateEthereumKeyContainer.tsx
```diff
@@ -0,0 +1,283 @@
...
+            } else {
+                errorsStore.setError({ error: e, expected: false });
+                setNextDisabled(true);
+                setLoading(false);
+                return;
+            }
+        }
+    }
+
+    async function onChangeWord(index: number, word: string) {
```
theblockstalk, line 130: It would be much better if we can turn the 6 word passphrase edit into a component that has a writeable set of 6 words as inputs and output. it should handle all the logic of verifying if the words are in the list and displaying errors if they arent.

the consuming container just needs to be able to tell it which passphrase words it should have and be able to read any changes.

src/containers/MainContainer.tsx
```diff
@@ -191,13 +193,55 @@ export default function MainContainer({ did }: { did?: string }) {
...
-
-                                    <Text style={styles.balanceView}>
-                                        {formatCurrencyValue(balance) || 0} LEOS
+                                    <View style={{ flexDirection: 'column', alignItems: 'flex-end' }}>
+                                        <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+                                            <Text> {formatCurrencyValue(balance) || 0} LEOS</Text>
+                                        </View>
                                         <Text style={styles.secondaryColor}>
-                                            (${balance ? formatCurrencyValue(balance * 0.002) : 0.0})
+                                            ${balance ? formatCurrencyValue(balance * 0.002) : 0.0}
```
theblockstalk, line 201: make 0.002 a const value somewhere....

src/services/CommunicationModule.ts
```diff
@@ -168,14 +169,33 @@ export default function CommunicationModule() {
...
+
+                        sendWalletConnectNotificationOnBackground(
+                            'Transaction Request',
+                            'Ethereum transaction signing request'
+                        );
+                        break;
+                    }
+
+                    default:
+                        return 'Method not supported';
```
theblockstalk, line 196: returning a string? is this a good idea?

src/components/PassphraseInput.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { useState, useEffect } from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+import AutoCompletePassphraseWord from '../components/AutoCompletePassphraseWord';
+import theme, { commonStyles } from '../utils/theme';
+import { util } from '@tonomy/tonomy-id-sdk';
+
+interface PassphraseInputProps {
+    initialPassphrase: string[];
```
theblockstalk, line 8: This would keep convention with typical React components much better:

```ts
value: string[];
onChange: (passphrase: string[]) => void;
```

src/components/PassphraseInput.tsx
theblockstalk, line 1: great component!

src/containers/LoginPassphraseContainer.tsx
```diff
@@ -8,11 +8,11 @@ import TInfoBox from '../components/TInfoBox';
 import LayoutComponent from '../components/layout';
 import { Props } from '../screens/LoginPassphraseScreen';
 import useUserStore, { UserStatus } from '../store/userStore';
-import { AccountType, SdkError, SdkErrors, TonomyUsername, util, TonomyContract } from '@tonomy/tonomy-id-sdk';
+import { AccountType, SdkError, SdkErrors, TonomyUsername, TonomyContract } from '@tonomy/tonomy-id-sdk';
 import { generatePrivateKeyFromPassword, savePrivateKeyToStorage } from '../utils/keys';
 import useErrorStore from '../store/errorStore';
 import { DEFAULT_DEV_PASSPHRASE_LIST } from '../store/passphraseStore';
-import AutoCompletePassphraseWord from '../components/AutoCompletePassphraseWord';
```
theblockstalk, line 15: can we delete `'../components/AutoCompletePassphraseWord';` now? I assume it is not used

src/utils/chain/etherum.ts
```diff
@@ -30,6 +30,8 @@ import { currentETHAddress, web3wallet } from '../../services/WalletConnect/Wall
 import { SessionTypes, SignClientTypes } from '@walletconnect/types';
 import { getSdkError } from '@walletconnect/utils';
 
+export const USD_CONVERSION = 0.002;
```
theblockstalk, line 33: we need to create a new PangeaMainnetChain and LEOSAsset. In these classes we can then set that the USD amount for LEOS is $0.012 / LEOS. Can be in future task

src/containers/LoginPassphraseContainer.tsx
sadiabbasi: we used this inside passphraseInput.tsx component

theblockstalk (approved):

PR #1031, opened 2024-07-07
Feature/987 implement account blance

sadiabbasi:


src/components/AccountDetails.tsx
```diff
@@ -1,48 +1,107 @@
...
+import { formatCurrencyValue } from '../utils/numbers';
+import { Images } from '../assets';
 
 export type AccountDetailsProps = {
-    accountName: string;
+    accountDetails: {
+        symbol: string;
+        name: string;
+        icon?: ImageSourcePropType | undefined;
+        ethBalance?: string;
```
theblockstalk, line 18: just `balance`. This could be other types of assets.

Why not just pass the Asset and Account object here instead? Might be simpler and then more obvious what the component is, but also not necessary.

src/containers/MainContainer.tsx
```diff
@@ -216,7 +239,16 @@ export default function MainContainer({
                                 </View>
                             </TouchableOpacity>
 
-                            <TouchableOpacity onPress={() => (refMessage.current as any)?.open()}>
+                            <TouchableOpacity
+                                onPress={() => {
+                                    setAccountDetails({
+                                        symbol: 'Eth',
+                                        icon: require('../assets/icons/eth-img.png'),
```
theblockstalk, line 246: this should come from `asset.getChain().getLogo()`, same with symbol, name, address...

theblockstalk (commented):
use our abstractions more

PR #1033, opened 2024-07-08
simplify hcpatch conditions

sadiabbasi:
I am not able to replicate the error but may be we can simplify conditions and check the code is settings or not 
reference: https://github.com/hCaptcha/react-native-hcaptcha/blob/HEAD/Example.App.js


sadiabbasi:
update hcaptcha in mastr in separte branch

PR #1079, opened 2024-07-31
Feature/178 smart contract transaction

sadiabbasi:


test/utils/ethereum.test.ts
```diff
@@ -56,4 +81,76 @@ describe('Ethereum sign transaction', () => {
...
     }, 30000);
+
+    it('deploy smart contract send raw transaction', async () => {
+        // replace this data with data from your blockchain
+        const senderAddress = '0xd2a010dcC01bF923C9aa329bA88F61428B61e4f2';
+        const senderPrivateKey = '0xe75d0d4c927c0848a6c56fc7f20850959157ba109a3c9b0e40818e8ef0bc4a63';
+        const nonce = await web3.eth.getTransactionCount(senderAddress, 'latest');
+        const gasPrice = await web3.eth.getGasPrice();
+        const gasLimit = 3000000;
+        const contractAddress = '0x781DE4eeD3B6f2aaAdB35335e597995E05b0A2d5';
```
sadiabbasi, line 92: @theblockstalk how i can make these address dynamic?

theblockstalk: the gnache / truffle framework has functions to deploy contract dynamically

sadiabbasi:
> a bit of tidying up to do like removing testing code, and comments
> 
> If I run `yarn run test` without anything else will it work?

you need to run ganache-cli then it works

package.json
```diff
@@ -18,7 +18,9 @@
         "start": "NODE_ENV=development yarn run build:prepare && NODE_ENV=development npx expo start --dev-client --clear",
         "lint": "eslint src test --ext '[.ts, .tsx]' --resolve-plugins-relative-to .",
         "lint:fix": "yarn run lint -- --fix",
-        "test": "jest",
+        "test:ganache-start": "pm2 start 'npx ganache-cli' --name ganache --silent",
+        "test:ganache-stop": "pm2 stop ganache && pm2 delete ganache",
+        "test": "yarn run test:ganache-start && jest && yarn run test:ganache-stop",
```
theblockstalk, line 23: if `jest` fails, then the cleanup `test:ganache-stop` will not run. could this be a problem?

package.json
```diff
@@ -56,6 +58,7 @@
         "expo-sqlite": "~11.3.3",
         "expo-status-bar": "~1.6.0",
         "fast-text-encoding": "^1.0.6",
+        "fs": "^0.0.1-security",
```
theblockstalk, line 61: shouldn't `fs` and `solc` be devDependancies?

.github/workflows/tests.yaml
```diff
@@ -9,16 +9,37 @@ jobs:
...
       - name: 🏗 Setup repo
         uses: actions/checkout@v4
 
-      - name: 📦 Build, test and lint
+      - name: 📦 Setup Node.js
         uses: actions/setup-node@v4
         with:
           node-version: 18.12.1
+          
+      - name: Wait for Ganache to be ready
```
theblockstalk, line 27: cleanup...

.github/workflows/tests.yaml
```diff
@@ -9,16 +9,37 @@ jobs:
   build:
     runs-on: ubuntu-20.04
 
+    services:
```
theblockstalk, line 12: now not needed?

src/utils/chain/etherum.ts
```diff
@@ -313,7 +315,18 @@ export class EthereumTransaction implements ITransaction {
         const decodedData = new Interface(abi).parseTransaction({ data: this.transaction.data });
 
         if (!decodedData?.args) throw new Error('Failed to decode function name');
-        return decodedData.args;
+
+        // Accessing the inputs of the fragment
```
theblockstalk, line 319: this comment is still not resolved https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1079/files#r1698386237

if no ABI is retrieved, then this function needs to tell the UI (error or return value) what to do - which is not not show the user the function or arguments, just a simplified UI.

sadiabbasi:
``` if (['contract', 'both'].includes(transactionType.toString())) {
                    try {
                        functionName = await transaction.getFunction();
                        args = await transaction.getArguments();
                        setDisplayFunction(true);
                    } catch (error) {
                        if (
                            error.message === 'Not a contract call' ||
                            error.message === 'Transaction has no data' ||
                            error.message === 'Failed to decode function name'
                        ) {
                            functionName = null;
                            args = null;
                        } else {
                            errorStore.setError({ error, expected: false });
                        }
                    }
                }``` implement this if no function found

Comment (theblockstalk) [reply src/utils/chain/etherum.ts:319]: I think it would be cleaner if `getFunction()` and `getArguments()` Would return null when there is no ABI available on etherscan.io

File: src/utils/chain/etherum.ts [hunk 7]
```diff
@@ -294,26 +295,46 @@ export class EthereumTransaction implements ITransaction {
...
+    async getArguments(): Promise<Record<string, string> | null> {
+        try {
+            const abi = await this.fetchAbi();
+
+            if (!this.transaction.data) throw new Error('Transaction has no data');
+            const decodedData = new Interface(abi).parseTransaction({ data: this.transaction.data });
 
-        if (!this.transaction.data) throw new Error('Transaction has no data');
-        const decodedData = new Interface(abi).parseTransaction({ data: this.transaction.data });
+            if (!decodedData?.args) throw new Error('Failed to decode function name');
```
theblockstalk, line 322: failed to decode function args...

src/utils/chain/etherum.ts
```diff
@@ -294,26 +295,46 @@ export class EthereumTransaction implements ITransaction {
...
-        if (!this.transaction.data) throw new Error('Transaction has no data');
-        const decodedData = new Interface(abi).parseTransaction({ data: this.transaction.data });
+            if (!this.transaction.data) throw new Error('Transaction has no data');
+            const decodedData = new Interface(abi).parseTransaction({ data: this.transaction.data });
 
-        if (!decodedData?.name) throw new Error('Failed to decode function name');
-        return decodedData.name;
+            if (!decodedData?.name) throw new Error('Failed to decode function name');
+            return decodedData.name;
+        } catch (error) {
```
theblockstalk, line 311: This will work but is not great error handling. Suppose a different kind of error happens, how I going to know about it?
`if (!decodedData?.name) return null` might be a better way to do this without blanketing all errors.

sadiabbasi:
Hopefully will fix in the upcoming PR https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1115

PR #1087, opened 2024-08-08
Development

sadiabbasi:


src/config/config.production.json
```diff
@@ -39,5 +39,9 @@
     "loggerLevel": "error",
     "captchaSiteKey": "0caf1023-edb9-466c-b848-6f2e0fe975eb",
     "blockExplorerUrl": "https://explorer.pangea.web4.world",
-    "tonomyIdSlug": "united-wallet://"
+    "tonomyIdSlug": "united-wallet://",
```
theblockstalk, line 42: Could you please make a Issue - All of the keys here should be last in by environment variable only. Having them hardcoded here is a security vulnerability

package.json
```diff
@@ -24,17 +24,25 @@
         "updateSdkVersion": "./update_sdk_version.sh"
     },
     "dependencies": {
-        "@hcaptcha/react-native-hcaptcha": "^1.5.1",
+        "@hcaptcha/react-native-hcaptcha": "1.7.2",
+        "@json-rpc-tools/utils": "^1.7.6",
```
theblockstalk, line 28: is this package really needed? seems like we could do this without an extra dependancy

package.json
```diff
@@ -53,19 +63,26 @@
         "react-native-autocomplete-input": "^5.3.2",
         "react-native-exception-handler": "^2.10.10",
         "react-native-gesture-handler": "~2.12.0",
+        "react-native-get-random-values": "^1.9.0",
```
theblockstalk, line 66: is there not an expo package for this?

PR #1115, opened 2024-08-13
Feature/1111 antelope chain abstraction

theblockstalk:


src/utils/chain/antelope.ts
```diff
@@ -0,0 +1,497 @@
...
+    PangeaTestnetChain,
+    'LEOS',
+    'LEOS',
+    6,
+    'https://github.com/Tonomy-Foundation/documentation/blob/master/images/logos/LEOS%20256x256.png?raw=true',
+    'ethereum'
+);
+
+PangeaMainnetChain.setNativeToken(LEOSToken);
+PangeaTestnetChain.setNativeToken(LEOSTestnetToken);
```
sadiabbasi, line 297: so this is not testable in staging

theblockstalk: should be possible

theblockstalk: we might need to also create a token for EOS and EOSJungleTestnet for your testing. Should be pretty easy.

src/utils/chain/etherum.ts
```diff
@@ -138,30 +115,25 @@ export class EthereumToken extends AbstractToken {
...
         const lookupAccount: IAccount =
             account ||
-            this.account ||
+            this.getAccount() ||
             (() => {
                 throw new Error('Account not found');
             })();
 
-        const balanceWei = await this.chain.getProvider().getBalance(lookupAccount.getName() || '');
+        const balanceWei = (await lookupAccount.getBalance(this.chain.getNativeToken())).getAmount();
```
sadiabbasi, line 136: this is calling infinity  ERROR  There was a problem sending log messages to your development environment [RangeError: Maximum call stack size exceeded (native stack depth)]
PayloadTooLargeError: request entity too large

theblockstalk: Perhaps it is creating a infinite loop. This can be tested by adding logging in `EthereumToken.getBalance()` `EthereumAccount.getBalance()` (it is probably between these two) `EthereumChain.getNativeToken()` and `EthereumToken.getAmount()`

PR #1116, opened 2024-08-18
Bug/1112 no internet app freeze

sadiabbasi:


src/store/errorStore.ts
```diff
@@ -24,8 +25,16 @@ const useErrorStore = create<ErrorState>((set, get) => ({
...
-    setError: ({ error, title, expected, onClose }) => {
+    setError: async ({ error, title = '', expected, onClose }) => {
         set({ error, title, expected, onClose });
+
+        const state = await NetInfo.fetch();
+
+        if (state.isConnected) {
+            set({ error, title, expected, onClose });
+        } else {
+            console.log('Network is off. Error will not be thrown.');
```
theblockstalk, line 36: use `debug` instead

src/utils/chain/types.ts
```diff
@@ -148,7 +148,13 @@ export abstract class AbstractAsset implements IAsset {
...
     }
     getSymbol(): string {
-        return this.token.getSymbol();
+        const symbol = this.token.getSymbol();
+
+        if (symbol) {
+            return symbol;
+        }
+
+        return '';
```
theblockstalk, line 157: this does not seem right

src/utils/chain/types.ts
```diff
@@ -164,7 +170,7 @@ export abstract class AbstractAsset implements IAsset {
     }
 
     toString(): string {
-        return `${this.printValue()} ${this.token.getSymbol()}`;
+        return `${this.printValue()} ${this.token?.getSymbol() || ''}`;
```
theblockstalk, line 173: this does not seem right. this will create issues later

theblockstalk (commented):
While this approach may work, it is blanketing the issue. The real issue is error handling of internet requests.

What happens when we create an Internet request when the internet is not connected? do we get a timeout, or a HTTP error code? let's handle that instead. Also we should give some feedback to the user - which we should get a design for.

Ideally, the user is not connected to the Internet the app should not fail it should properly just stay in the same state. A long-term solution ( we should not implement this yet) would be to have some kind of queue of Internet requests that should stack up while the user uses the app and are all sent when the Internet connection returns...

src/components/TErrorModal.tsx
```diff
@@ -27,11 +27,13 @@ export default function TErrorModal(props: TErrorModalProps) {
...
     }
 
-    if (props?.expected === false) {
-        console.error(props.error);
-        debug(JSON.stringify(props.error, null, 2), props.expected);
-        // TODO: log to Tonomy Foundation team
-    }
+    useEffect(() => {
+        if (props.expected === false) {
+            console.error('Error Modal:', props.error, props?.error?.message);
```
theblockstalk, line 32: `props?.error?.message` unnecesarry

src/containers/MainSplashContainer.tsx
```diff
@@ -10,58 +10,72 @@ import { SdkError, SdkErrors } from '@tonomy/tonomy-id-sdk';
...
-
-        // Cleanup the event listener
-        return () => unsubscribe();
-    }, [navigation]);
-
     useEffect(() => {
         async function main() {
             await sleep(800);
 
+            // try {
```
theblockstalk, line 24: remove comment

src/store/userStore.ts
```diff
@@ -48,6 +49,7 @@ const useUserStore = create<UserState>((set, get) => ({
     // @ts-ignore PublicKey type error
     user: createUserObject(new RNKeyManager(), storageFactory),
     status: UserStatus.NONE,
+    isUserInitialized: false,
```
theblockstalk, line 52: name is a bit misleading as this is about the app being initialized, not the user. `isAppInitialized` would be better

src/containers/MainSplashContainer.tsx
```diff
@@ -10,58 +10,72 @@ import { SdkError, SdkErrors } from '@tonomy/tonomy-id-sdk';
...
-                    case UserStatus.LOGGED_IN:
-                        try {
-                            await user.getUsername();
-                            await initializeWalletState();
-                        } catch (e) {
-                            if (e instanceof SdkError && e.code === SdkErrors.InvalidData) {
-                                logout("Invalid data in user's storage");
-                                clearState();
-                            } else {
-                                throw e;
```
theblockstalk, line 58: what about this condition? seems important that we handle all errors

src/containers/MainSplashContainer.tsx
```diff
@@ -10,58 +10,72 @@ import { SdkError, SdkErrors } from '@tonomy/tonomy-id-sdk';
...
             try {
-                await initializeStatusFromStorage();
-                const status = getStatus();
+                if (!isUserInitialized) await initializeStatusFromStorage();
 
                 await connect();
 
-                switch (status) {
-                    case UserStatus.NONE:
-                        navigation.dispatch(StackActions.replace('Home'));
```
theblockstalk, line 44: `NONE` and `NOT_LOGGED_IN` are not handled... Is this keeping the expected logic?

src/store/errorStore.ts
```diff
@@ -24,7 +24,7 @@ const useErrorStore = create<ErrorState>((set, get) => ({
     title: undefined,
     expected: undefined,
     onClose: undefined,
-    setError: ({ error, title, expected, onClose }) => {
+    setError: async ({ error, title = '', expected, onClose }) => {
```
theblockstalk, line 27: blank title? Modal should always have a title IMO. At least "something went wrong"

src/store/useWalletStore.ts
```diff
@@ -75,8 +75,8 @@ const useWalletStore = create<WalletState>((set, get) => ({
                     return {
                         account,
                         balance: {
-                            balance: balance.toString(),
-                            usdBalance: await balance.getUsdValue(),
+                            balance: balance?.toString() || '0',
```
theblockstalk, line 78: A better strategy would be to show a spinning wheel to give user feedback that this is being requested. Let's implement something like this when we do the Wallet UI upgrade. Please ensure Pavel has sorted such transitions.

src/utils/chain/types.ts
```diff
@@ -164,7 +164,7 @@ export abstract class AbstractAsset implements IAsset {
     }
 
     toString(): string {
-        return `${this.printValue()} ${this.token.getSymbol()}`;
+        return `${this.printValue()} ${this.token?.getSymbol()}`;
```
theblockstalk, line 167: token may be undefined? then maybe should return an error because it does not make sense to print the string without asset symbol.

src/store/errorStore.ts
sadiabbasi: at some places we are not setting the title i will update this

src/navigation/Root.tsx
```diff
@@ -33,6 +33,9 @@ import SignTransactionConsentSuccessScreen from '../screens/SignTransactionConse
 import WalletConnectLoginScreen from '../screens/WalletConnectLoginScreen';
 import CreateEthereumKeyScreen from '../screens/CreateEthereumKeyScreen';
 import { IPrivateKey, ISession, ITransaction, TransactionType } from '../utils/chain/types';
+import Debug from 'debug';
+
+const debug = Debug('tonomy-id:store:root-navigation');
```
theblockstalk, line 38: `tonomy-id:navigation:root` is consistent

src/containers/MainSplashContainer.tsx
```diff
@@ -10,38 +10,30 @@ import { SdkError, SdkErrors } from '@tonomy/tonomy-id-sdk';
...
-
     useEffect(() => {
         async function main() {
             await sleep(800);
 
             try {
-                await initializeStatusFromStorage();
-                const status = getStatus();
+                if (!isAppInitialized) {
+                    await initializeStatusFromStorage();
```
theblockstalk, line 27: is the error for this now caught correctly??? I don't see new error handling in this component - but perhaps it is in a different component...

Also: we can better handle the network error here:
```
if (!isAppInitialized) {
  // while initializeStatusFromStorage(); throws a network error, then keep looping. if other error thrown then exit loop
  // loop progressively slower starting at 10s interval but doubling till interval = 1hr maximum
     await initializeStatusFromStorage();
}
```
this will then implement a progresses re-attempt of the initialization so that the user islogged out if the log out on a different device

src/containers/MainSplashContainer.tsx
```diff
@@ -10,38 +10,45 @@ import { SdkError, SdkErrors } from '@tonomy/tonomy-id-sdk';
...
+                    let retryInterval = 10000; // Start with 10 seconds
+                    const maxInterval = 3600000; // Cap at 1 hour
+
+                    while (!isAppInitialized) {
+                        try {
+                            await initializeStatusFromStorage();
+                            break;
+                        } catch (error) {
+                            if (error.message === 'Network request failed') {
+                                console.error('Network error occurred, retrying in', retryInterval / 1000, 'seconds');
```
theblockstalk, line 35: use a debug log here instead...

theblockstalk: as the user has not done anything intentionally wrong we should not `console.error` this.

PR #1119, opened 2024-08-20
Add Bottom Navigation

jekingohel:


src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+type ScanTabBarButtonProps = BottomTabBarButtonProps;
+const ScanTabBarButton: React.FC<ScanTabBarButtonProps> = ({ children, onPress }) => (
+    <TouchableOpacity style={{ top: -30, justifyContent: 'center', alignContent: 'center' }} onPress={onPress}>
+        <View
+            style={{
+                width: 70,
+                height: 70,
+                borderRadius: 35,
+                backgroundColor: '#000000',
```
sadiabbasi, line 37: use color from theme

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,154 @@
...
+    Explore: undefined;
+    Scan: undefined;
+    Apps: undefined;
+};
+
+const Tab = createBottomTabNavigator<RouteStackParamList>();
+
+type ScanTabBarButtonProps = BottomTabBarButtonProps;
+const ScanTabBarButton: React.FC<ScanTabBarButtonProps> = ({ children, onPress }) => (
+    <TouchableOpacity style={{ top: -30, justifyContent: 'center', alignContent: 'center' }} onPress={onPress}>
```
sadiabbasi, line 31: don't recomend to add negative top?

jekingohel: As per figma design, QR scan button slightly top to the bottom tab. So its required to use negative top value.

src/components/CustomDrawer.tsx
```diff
@@ -12,6 +12,7 @@ import useWalletStore from '../store/useWalletStore';
 // https://callstack.github.io/react-native-paper/docs/guides/icons/
 const icons: Record<keyof DrawerStackParamList, string> = {
     UserHome: 'home',
+    BottomTabs: 'home',
```
sadiabbasi, line 15: better to change the value to some thing related to bottomTabs.. may be bottomMenuBar or bottomNavigation something like this?

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,154 @@
...
+        <Tab.Navigator
+            screenOptions={{
+                tabBarActiveTintColor: theme.colors.primary,
+                tabBarInactiveTintColor: theme.colors.grey,
+                headerShown: false, // or true if you want headers on tabs
+            }}
+        >
+            <Tab.Screen
+                name="UserHome"
+                component={MainScreen}
```
sadiabbasi, line 60: for the citizenship tab we will use the citizen component and name not use the userhome

src/navigation/Root.tsx
```diff
@@ -203,7 +204,7 @@ export default function RootNavigation() {
                 <>
                     <NotificationModule />
                     <CommunicationModule />
-                    <Stack.Navigator initialRouteName={'UserHome'} screenOptions={defaultScreenOptions}>
+                    <Stack.Navigator initialRouteName={'BottomTabs'} screenOptions={defaultScreenOptions}>
```
sadiabbasi, line 207: with this change after login which screen is display by default?

jekingohel: After login it will display Citizenship screen.

src/navigation/BottomTabNavigator.tsx
jekingohel: Its updated on Citizenship PR changes.

src/components/CustomDrawer.tsx
jekingohel: update to BottomNavigation

src/navigation/BottomTabNavigator.tsx
sadiabbasi: I still thinks the same so when we merge this PR aand from the bottom menu when click on scan icon it should open the mainScreen not the citizenship page

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,154 @@
...
+                options={{
+                    tabBarLabel: ({ focused }) => (
+                        <Text
+                            style={[styles.tabLabel, { color: !focused ? theme.colors.tabGray : theme.colors.black }]}
+                        >
+                            Citizenship
+                        </Text>
```
sadiabbasi, line 67: this label text repeated may be create a function that return this view and take arguments (label, focused)

src/navigation/BottomTabNavigator.tsx
sadiabbasi: and the default page open after just this PR merge will be main screen @theblockstalk am i ritght?

PR #1121, opened 2024-09-02
Feature/1061 esr transaction

sadiabbasi:


src/containers/MainContainer.tsx
```diff
@@ -141,6 +158,92 @@ export default function MainContainer({
         try {
             if (data.startsWith('wc:')) {
                 if (web3wallet) await web3wallet.core.pairing.pair({ uri: data });
+            } else if (data.startsWith('esr:')) {
+                let chain: AntelopeChain;
```
sadiabbasi, line 162: how we can listen the antelope event any guide on this or currently it will be on scan qr code?

src/containers/MainContainer.tsx
```diff
@@ -141,6 +158,92 @@ export default function MainContainer({
...
+
+                if (settings.env === 'testnet') {
+                    chain = PangeaTestnetChain;
```
theblockstalk, line 165: I think the QR code data will contain a chainId, Or some other data that tells us which Antelope chain to use. Based on this we will use the appropriate Chain object.

src/containers/MainContainer.tsx
theblockstalk: You mean how do we listen for the transaction event after the sessions is created? I'm not sure. Might have to look at the docks and how they do this or look the Anchore desktop implementation which does this.

src/utils/chain/types.ts
```diff
@@ -310,16 +310,17 @@ export class Asset extends AbstractAsset {
...
-    id: number;
-    topic: string;
+export interface IChainSession {
+    createSession(request: unknown): Promise<void>;
+    cancelLoginRequest(request: unknown): Promise<void>;
+    disconnectSession(): Promise<void>;
+    createTransactionRequest(request: unknown): Promise<unknown>;
+    approveTransactionRequest(request: unknown, transaction?: unknown): Promise<void>;
+    rejectTransactionRequest(request: unknown): Promise<void>;
+    getActiveAccounts(): Promise<ChainDetail[]>;
```
theblockstalk, line 325: why not use `IAccount` here?

it contains the account name + the chain which contains the network name and chainId

src/utils/chain/types.ts
```diff
@@ -310,16 +310,17 @@ export class Asset extends AbstractAsset {
...
+    networkName: string;
 }
-
-export interface ISession {
-    origin: string;
-    id: number;
-    topic: string;
+export interface IChainSession {
+    createSession(request: unknown): Promise<void>;
+    cancelLoginRequest(request: unknown): Promise<void>;
```
theblockstalk, line 320: maybe `cancelSessionRequest` Would be better so that we know that this is about the session. Adding new terminology "login" would be confusing

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
-    constructor(payload: SignClientTypes.EventArguments['session_proposal'], chain: EthereumChain) {
-        this.payload = payload;
-        this.chain = chain;
+    constructor(wallet: IWeb3Wallet) {
+        this.wallet = wallet;
     }
 
-    getId(): number {
-        return this.payload.id;
+    public setNamespaces(namespaces: SessionTypes.Namespaces): void {
```
theblockstalk, line 424: You don't need to use the `public` keyword - everything in JS is public unless explicitly declared otherwise

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
-            return this.payload.params.proposer.metadata.icons[0];
-        }
+    async getActiveAccounts(): Promise<ChainDetail[]> {
+        return this.chainAccountList;
+    }
+
+    async createSession(request: SignClientTypes.EventArguments['session_proposal']): Promise<void> {
+        await this.wallet?.approveSession({
+            id: request.id,
+            relayProtocol: request.params.relays[0].protocol,
```
theblockstalk, line 443: Will this only works for the 1st protocol specified? What if the request wants to log into several chains?

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
+            namespaces: this.getNamespaces(),
+        });
+    }
+
+    async cancelLoginRequest(request: SignClientTypes.EventArguments['session_proposal']): Promise<void> {
+        await this.wallet?.rejectSession({
```
theblockstalk, line 449: Should throw error if this.wallet is undefined

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
+            id: request.id,
+            reason: getSdkError('USER_REJECTED'),
+        });
+    }
+
+    async disconnectSession(): Promise<void> {
+        // Logic to disconnect the WalletConnect session
```
theblockstalk, line 456: TODO

src/utils/chain/types.ts
```diff
@@ -310,16 +310,17 @@ export class Asset extends AbstractAsset {
...
+    disconnectSession(): Promise<void>;
+    createTransactionRequest(request: unknown): Promise<unknown>;
```
theblockstalk, line 322: `request: ITransaction` is this a good idea here?

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
+        // Example: Disconnect WalletConnect provider
+    }
+
+    async createTransactionRequest(transaction: ITransaction): Promise<TransactionRequest> {
```
theblockstalk, line 460: Because this is the EthereumSession class, you can use `EthereumTransaction` here instead of `ITransaction` which will remove the type overload on L465 and make it clearer for typescript and developers

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
+        const transactionRequest: TransactionRequest = {
+            to: (await transaction.getTo()).getName(),
+            from: (await transaction.getFrom()).getName(),
+            value: ethers.parseEther(parseFloat((await transaction.getValue()).toString()).toFixed(18)),
```
theblockstalk, line 464: can't you use transaction.getValue().getAmount() perhaps?

src/utils/chain/etherum.ts
```diff
@@ -410,32 +412,81 @@ export class EthereumAccount extends AbstractAccount {
...
+            data: await (transaction as EthereumTransaction).getData(),
+        };
+
+        return transactionRequest;
+    }
+
+    async approveTransactionRequest(
+        request: Web3WalletTypes.SessionRequest,
+        signedTransaction: unknown
```
theblockstalk, line 473: missing type here

src/utils/chain/antelope.ts
```diff
@@ -89,20 +95,28 @@ export class AntelopePrivateKey extends AbstractPrivateKey implements IPrivateKe
...
-            const abi = abiMap.get(data.account.toString());
-
-            actionData.push(Action.from(data, abi));
-        });
+        for (const action of actions) {
+            if (action.name.toString() === 'identity') {
+                const abi: ABI = Serializer.synthesize(IdentityV3 as ABISerializableConstructor);
+
+                // eslint-disable-next-line camelcase
+                abi.actions = [{ name: 'identity', type: 'identity', ricardian_contract: '' }];
```
theblockstalk, line 107: why?

src/utils/chain/antelope.ts
```diff
@@ -89,20 +95,28 @@ export class AntelopePrivateKey extends AbstractPrivateKey implements IPrivateKe
...
+                actionData.push(Action.from(action, abi));
+            } else {
+                const api = await this.chain.getApiClient();
+                const uniqueContracts = [...new Set(actions.map((data) => data.account))];
```
theblockstalk, line 111: Why did you move the abiArray generation inside the loop? this is only needed to be done once before the loop starts AFAIKT

src/utils/chain/antelope.ts
```diff
@@ -213,14 +228,20 @@ export class AntelopeToken extends AbstractToken implements IToken {
...
-            case 'Pangea Testnet':
-                return LEOS_PUBLIC_SALE_PRICE;
-            default:
-                throw new Error('Unsupported Antelope chain');
-        }
+        console.log('his.getChain()', this.getChain());
+        //TODO check getChain is undefined coming from antelope chain PR
+        return LEOS_PUBLIC_SALE_PRICE;
+
+        // switch (this.getChain().getName()) {
```
theblockstalk, line 235: leave the switch statement. Each chain may have different price and this allows us to set that properly.

src/utils/chain/antelope.ts
```diff
@@ -310,8 +347,9 @@ export class AntelopeAction implements IOperation {
     async getType(): Promise<TransactionType> {
         // TODO: need to also handle token transfers on other contracts
         if (
-            this.action.name.toString() === 'transfer' &&
-            this.action.account.toString() === this.chain.getNativeToken().getContractAccount()?.getName().toString()
```
theblockstalk, line 314: this line Should not be deleted. Otherwise it could be a notification transaction ( not the actual transfer)

src/utils/chain/antelope.ts
```diff
@@ -396,8 +435,19 @@ export class AntelopeTransaction implements ITransaction {
         const operationAmountsPromises = (await this.getOperations()).map(async (operation) =>
             (await operation.getValue()).getAmount()
         );
+
         const operationAmounts = (await Promise.all(operationAmountsPromises)).reduce((a, b) => a + b, BigInt(0));
-        const amount = operationAmounts + (await this.estimateTransactionFee()).getAmount();
+        let amount = operationAmounts + (await this.estimateTransactionFee()).getAmount();
+
+        if (typeof amount === 'string') {
```
theblockstalk, line 442: This is a patch. Better to ensure that the `operationAmounts` and `amount` are always a BigInt

src/utils/chain/antelope.ts
```diff
@@ -495,3 +545,79 @@ export class AntelopeAccount extends AbstractAccount implements IAccount {
...
+            },
+        ];
+    }
+
+    async disconnectSession(): Promise<void> {
+        // Logic to disconnect the ESR session
+        // Example: Clear the session and any stored data
+    }
+
+    async createTransactionRequest(transaction: ITransaction): Promise<ITransaction> {
```
theblockstalk, line 581: can use `AntelopeTransaction` for better type safety and developer understanding

src/utils/chain/antelope.ts
```diff
@@ -495,3 +545,79 @@ export class AntelopeAccount extends AbstractAccount implements IAccount {
...
+    private account: AntelopeAccount;
+    private antelopeKey: AntelopePrivateKey;
+    private chain: AntelopeChain;
+
+    constructor(antelopeKey: AntelopePrivateKey, chain: AntelopeChain) {
+        this.antelopeKey = antelopeKey;
+        this.chain = chain;
+    }
+
+    async createSession(request: ResolvedSigningRequest): Promise<void> {
```
theblockstalk, line 560: why is this not implemented/used?

src/store/userStore.ts
```diff
@@ -92,17 +87,18 @@ async function printStorage(message: string) {
...
     const secureKeys: string[] = [];
 
     for (const level of Object.keys(KeyManagerLevel)) {
+        console.log(KEY_STORAGE_NAMESPACE + level);
         const value = await SecureStore.getItemAsync(KEY_STORAGE_NAMESPACE + level);
 
         if (value) secureKeys.push(KEY_STORAGE_NAMESPACE + level);
     }
 
-    debug(message, 'SecureStore keys', secureKeys);
```
theblockstalk, line 105: please use `debug` do not change to `console.log`.

You can view logs by setting the `DEBUG` env variable. e.g. `export DEBUG=tonomy-id:store:userStore` will show you these debug logs

src/store/userStore.ts
```diff
@@ -1,4 +1,4 @@
-import { create } from 'zustand';
```
theblockstalk, line 1: pretty sure this is the correct way for zustand now. `import create from ...` is depreciated.

src/store/useWalletStore.ts
```diff
@@ -53,35 +53,47 @@ const defaultState = {
...
+                    if (!key) {
+                        return null;
+                    }
 
-                if (key) {
                     const exportPrivateKey = await key.exportPrivateKey();
                     const privateKey = new EthereumPrivateKey(exportPrivateKey, chain);
                     const account = await EthereumAccount.fromPublicKey(chain, await privateKey.getPublicKey());
-                    const balance = await token.getBalance(account);
+                    //TODO uncomment after fix the antelope chain PR
```
theblockstalk, line 83: see https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1115/files#r1737129893

src/store/useWalletStore.ts
```diff
@@ -53,35 +53,47 @@ const defaultState = {
...
                 debug('Already initialized');
                 return;
             }
 
+            debug('Starting state initialization');
             const state = get();
+
+            debug('Current state:', state);
+
             const fetchAccountData = async (chain: EthereumChain, token: EthereumToken, keyName: string) => {
```
theblockstalk, line 72: add a return type for better type protection

src/services/CommunicationModule.ts
```diff
@@ -153,107 +155,223 @@ export default function CommunicationModule() {
...
+    const { ethereumAccount, sepoliaAccount, polygonAccount } = useWalletStore();
+
+    const getChainDetail = useCallback(
+        (chainIds) => {
+            const chainDetails = chainIds?.map((chainId) => {
+                let address;
+                let networkName;
+
+                if (chainId === '11155111') {
+                    address = sepoliaAccount ? sepoliaAccount.getName() : '';
```
theblockstalk, line 167: if they are asking for these accounts and we do not have it in storage, that means we have not created them yet. We cannot send an empty string back or it may break the app. We need to send the user to the passphrase screen to generate the accounts first, then get them to finish this flow.

src/navigation/Root.tsx
```diff
@@ -65,7 +67,9 @@ export type RouteStackParamList = {
     SignTransaction: {
         transaction: ITransaction;
         privateKey: IPrivateKey;
-        session: ISession;
+        origin: string;
+        request: Web3WalletTypes.SessionRequest | ResolvedSigningRequest;
```
theblockstalk, line 71: The transaction signing request should be provided through the IChainSession object right? The exception to this is when a transaction is scanned directly (only on Antelope)

src/containers/WalletConnectLoginContainer.tsx
```diff
@@ -117,14 +73,14 @@ export default function WalletConnectLoginContainer({
...
+                        {accounts?.map(({ networkName, address }, index) => (
                             <View style={styles.networkHeading} key={index}>
                                 <Image source={require('../assets/icons/eth-img.png')} style={styles.imageStyle} />
                                 <Text style={styles.nameText}>{networkName} Network:</Text>
-                                {currentETHAddress && (
+                                {address && (
                                     <Text style={commonStyles.textAlignCenter}>
-                                        {currentETHAddress.substring(0, 9)}....
-                                        {currentETHAddress.substring(currentETHAddress.length - 8)}
+                                        {address.substring(0, 9)}....
```
theblockstalk, line 82: Isn't there a function in the chain abstractions like `getShortName()` - we should use this here

src/containers/WalletConnectLoginContainer.tsx
theblockstalk, line 1: 😄 great abstraction use here!

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -50,6 +49,7 @@ export default function SignTransactionConsentContainer({
         usdFee: number;
         total: string;
         usdTotal: number;
+        actions: any;
```
theblockstalk, line 52: use type here

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -94,34 +115,32 @@ export default function SignTransactionConsentContainer({
...
                 const args: Record<string, string> | null = null;
-                const transactionValue = ethers.parseEther(parseFloat(value).toFixed(18));
-                const etherFee = ethers.parseEther(fee);
 
-                const totalTransactionCost = transactionValue + etherFee;
-                const balance = (
-                    await (await transaction.getFrom()).getBalance(transaction.getChain().getNativeToken())
-                ).getAmount();
+                if (value) {
+                    const transactionValue = ethers.parseEther(parseFloat(value).toFixed(18));
```
theblockstalk, line 129: This is ethereum specic. We want this to be covered in the abstractions

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -62,28 +62,49 @@ export default function SignTransactionConsentContainer({
...
-                const fromAccount = (await transaction.getFrom()).getName();
+                const actions = await transaction.getData();
+
+                let fromAccount, toAccount, value, usdValue, balance, transactionType;
+
+                if (Array.isArray(actions)) {
+                    const accountName = (await user.getAccountName()).toString();
 
-                const toAccount = (await transaction.getTo()).getName();
+                    fromAccount = accountName;
```
theblockstalk, line 90: If there is an array of operations, is this displayed in the UI?

What about smart contract transactions ( not transfers). Are these currently supported? this will be needed for Hypha

src/containers/MainContainer.tsx
```diff
@@ -141,6 +158,92 @@ export default function MainContainer({
...
+                        name: action.name.toString(),
+                        authorization: action.authorization.map((auth) => ({
+                            actor: auth.actor.toString(),
+                            permission: auth.permission.toString(),
+                        })),
+                        data: action.data,
+                    })
+                );
+                const actions = createAssetAction;
+                const privateKeyValue = privateKey || '';
```
theblockstalk, line 208: `|| ''` should we do this? if there is no private key then better throw an error?

src/containers/MainContainer.tsx
```diff
@@ -141,6 +158,92 @@ export default function MainContainer({
...
+                if (!isIdentity) {
+                    navigation.navigate('SignTransaction', {
+                        transaction,
+                        privateKey: antelopeKey,
+                        origin: chain.getApiOrigin(),
+                        request: resolvedSigningRequest,
+                        session,
+                    });
+                } else {
+                    const signedTransaction = await antelopeKey.signTransaction(transaction);
```
theblockstalk, line 222: need to send to transaction consent screen here

theblockstalk: TODO later...

src/containers/MainContainer.tsx
```diff
@@ -141,6 +158,92 @@ export default function MainContainer({
...
+                const privateKeyValue = privateKey || '';
+                const transaction = AntelopeTransaction.fromActions(actions, EOSJungleChain);
+                const antelopeKey = new AntelopePrivateKey(PrivateKey.from(privateKeyValue), EOSJungleChain);
+                const session = new ESRSession(antelopeKey, chain);
+
+                if (!isIdentity) {
+                    navigation.navigate('SignTransaction', {
+                        transaction,
+                        privateKey: antelopeKey,
+                        origin: chain.getApiOrigin(),
```
theblockstalk, line 217: we need to show the app origin, not the chain origin

src/containers/SignTransactionConsentContainer.tsx
theblockstalk: done in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/containers/SignTransactionConsentContainer.tsx
theblockstalk: done in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/containers/SignTransactionConsentContainer.tsx
theblockstalk: done in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/containers/MainContainer.tsx
theblockstalk: done in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/utils/chain/types.ts
theblockstalk: resolved in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/utils/chain/etherum.ts
theblockstalk: done in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/containers/MainContainer.tsx
theblockstalk: resolved in https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1147

src/utils/chain/types.ts
sadiabbasi: updated..

src/utils/chain/etherum.ts
sadiabbasi: it's not required... i just remove this..

src/containers/MainContainer.tsx
sadiabbasi: resolved

src/utils/chain/etherum.ts
sadiabbasi: fixed

src/utils/chain/types.ts
sadiabbasi: added type

src/utils/chain/etherum.ts
sadiabbasi: done

PR #1137, opened 2024-09-06
Bug/1112 no internet app freeze

sadiabbasi:


src/containers/MainContainer.tsx
```diff
@@ -79,12 +79,23 @@ export default function MainContainer({
     useEffect(() => {
         const initializeAndFetchBalances = async () => {
             if (!initialized && ethereumAccount && sepoliaAccount && polygonAccount) {
-                await initializeWalletState();
+                try {
+                    await initializeWalletState();
+                } catch (error) {
+                    if (error.message === 'Network request failed') {
```
theblockstalk, line 85: Better to put this on a progressively increasing re-try (like we did with initializeUser).

We should show the user something back while we are trying to fetch the balance. But that can be done in the Wallet UI upgrade. First priority is getting this fix out.

theblockstalk: Seems like we need this in several places. Suggest you make a helper function to do this

```ts
async function progressiveRetryOnNetworkError(fn: async function) {
  // do/while loop that keeps trying to execute the function with progressively longer waits
}

...
progressiveRetryOnNetworkError(async () => await initializeWalletState());
```

src/containers/MainContainer.tsx
```diff
@@ -206,24 +213,42 @@ export default function MainContainer({
...
-            const accountToken = await account.getNativeToken();
-            const logoUrl = accountToken.getLogoUrl();
-
-            setAccountDetails({
-                symbol: accountToken.getSymbol(),
-                name: capitalizeFirstLetter(account.getChain().getName()),
-                address: account?.getName() || '',
-                ...(logoUrl && { image: logoUrl }),
-            });
+        try {
```
theblockstalk, line 216: you can use 
```
progressiveRetryOnNetworkError(async () => {
if (account) {
                const accountToken = await account.getNativeToken();
                const logoUrl = accountToken.getLogoUrl();

                setAccountDetails({
                    symbol: accountToken.getSymbol(),
                    name: capitalizeFirstLetter(account.getChain().getName()),
                    address: account?.getName() || '',
                    ...(logoUrl && { image: logoUrl }),
                });
            }
        }

File: src/utils/helper.ts [hunk 3]
```diff
@@ -17,3 +20,28 @@ export function extractHostname(url): string {
...
 };
+
+export async function progressiveRetryOnNetworkError(fn, initialDelay = 10000, maxDelay = 3600000) {
+    let delay = initialDelay;
+    const condition = true;
+
+    while (condition) {
+        try {
+            await fn();
+            debug('App initialized successfully.');
```
theblockstalk, line 31: "Function called successfully" (this can be used for other functions that app initialization)

src/containers/MainContainer.tsx
```diff
@@ -206,24 +213,42 @@ export default function MainContainer({
...
         }
     };
 
     const onRefresh = React.useCallback(async () => {
-        setRefreshing(true);
-        updateBalance();
-        setRefreshing(false);
-    }, [updateBalance]);
+        try {
+            setRefreshing(true);
```
theblockstalk, line 242: you probably want `setRefreshing(true/false)` to be inside the `updateBalance()` so that it tells the user when it is refreshing, not for the whole period it tries to refresh

PR #1144, opened 2024-09-09
bug/1140-asset-storage

sadiabbasi:


src/utils/StorageManager/repositories/assetStorageManager.ts
```diff
@@ -35,4 +43,29 @@ export abstract class AssetStorageManager {
     public async deleteAll(): Promise<void> {
         await this.repository.deleteAll();
     }
+
+    public async emplaceAccountName(name: string, value: IAccount): Promise<void> {
```
theblockstalk, line 47: instead of `name: string` would be easier and safer to use `chain: IChain` and then use chain.getName() as the key

Also on L59

src/utils/StorageManager/repositories/assetStorageManager.ts
```diff
@@ -35,4 +43,29 @@ export abstract class AssetStorageManager {
...
+            existingAccount.value = value.getName();
+            existingAccount.updatedAt = new Date();
+            await this.repository.updateAccountName(existingAccount);
+        } else {
+            await this.repository.storeAccountName(name, value.getName());
+        }
+    }
+
+    public async findAccountByName(name: string): Promise<IAccount | null> {
+        const account = await this.repository.findAccountByName(name);
```
theblockstalk, line 60: this returns a object, not an `IAccount`. probably want to fix this so that we use the correct IAccount object

src/utils/StorageManager/entities/assetStorage.ts
```diff
@@ -1,9 +1,9 @@
 /* eslint-disable indent */
 import { Entity, Unique, PrimaryGeneratedColumn, Column, Index } from 'typeorm';
 
-@Entity('AssetStorage')
-@Unique(['name'])
-export class AssetStorage {
+@Entity('AssetsStorage')
```
theblockstalk, line 4: convention is that this would be singular I think

src/utils/StorageManager/repositories/assetStorageRepository.ts
```diff
@@ -1,17 +1,21 @@
...
     }
 
-    public async storeAccountBalance(name: string, value: string): Promise<AssetStorage> {
+    public async createAsset(assetName: string, accountName: string): Promise<IAssetStorage> {
         const assetStorageEntity = this.ormRepository.create({
-            name,
-            value,
+            assetName,
+            accountName,
+            balance: JSON.stringify({
```
theblockstalk, line 15: better to have separate colums for balance and usdBalance, than store JSON objects

src/utils/StorageManager/entities/assetStorage.ts
```diff
@@ -1,21 +1,25 @@
 /* eslint-disable indent */
 import { Entity, Unique, PrimaryGeneratedColumn, Column, Index } from 'typeorm';
 
-@Entity('AssetStorage')
-@Unique(['name'])
-export class AssetStorage {
+@Entity('IAssetStorage')
+@Unique(['assetName'])
+export class IAssetStorage {
```
theblockstalk, line 6: `AssetStorage` not `IAssetStorage` (the "I" Is indicating an interface which this is not - it is a storage type)

src/utils/StorageManager/repositories/assetStorageManager.ts
```diff
@@ -1,35 +1,50 @@
...
+    balance: Balance;
+}
+
 export abstract class AssetStorageManager {
     protected repository: AssetStorageRepository;
 
     constructor(repository: AssetStorageRepository) {
         this.repository = repository;
     }
+    public async createAsset(chain: EthereumChain, value: IAccount): Promise<void> {
```
theblockstalk, line 21: `chain: EthereumChain` --> `token: IToken` is better. otherwise you cannot find assets on the same chain.

Then you call it like `createAsset(ETHToken, ...)`

similar for L26, L39

PR #1162, opened 2024-09-18
Bug/status issue

sadiabbasi:


src/containers/MainSplashContainer.tsx
```diff
@@ -30,10 +30,13 @@ export default function MainSplashScreenContainer({ navigation }: { navigation:
             try {
                 if (!isAppInitialized) {
                     try {
-                        progressiveRetryOnNetworkError(async () => await initializeStatusFromStorage());
+                        await initializeStatusFromStorage();
```
theblockstalk, line 33: can we not make it, to keep the function consumer code simpler?
`await progressiveRetroyOnNetworkError(...)`

src/store/userStore.ts
theblockstalk, line 1: we need to look here and think about this design https://github.com/Tonomy-Foundation/Tonomy-ID/compare/0.112.0...development#diff-06a91a0f9d93ddbd14d70a2b1e44588b5d665f3bcc28cf731051f1abb0248f62

src/store/userStore.ts
```diff
@@ -51,15 +51,10 @@ const useUserStore = create<UserState>((set, get) => ({
     status: UserStatus.NONE,
     isAppInitialized: false,
     getStatus: async () => {
-        const status = await AsyncStorage.getItem(STORAGE_NAMESPACE + 'store.status');
-        const userstatus = get().status;
-
-        get().setStatus(status as UserStatus);
-
-        return status as UserStatus;
+        return get().status;
```
theblockstalk, line 54:
```ts
const storageStatus = await AsyncStorage.getItem(STORAGE_NAMESPACE + 'store.status');
if (storageStatus) {
        set({status: storageStatus});
        return storageStatus  as UserStatus;
} else {
        const stateStatus = get().status;
        get().setStatus(stateStatus);
        return stateStatus;
}

File: src/store/userStore.ts [hunk 3]
```diff
@@ -94,7 +89,7 @@ const useUserStore = create<UserState>((set, get) => ({
 
                 debug('network error condition status', status);
 
-                get().setStatus(status as UserStatus);
+                get().setStatus(UserStatus.LOGGED_IN);
 
                 throw e;
             } else {
```
theblockstalk, line 95: set({ isAppInitialized: true }) on new L98 (after the try/catch)

src/store/userStore.ts
```diff
@@ -94,7 +89,7 @@ const useUserStore = create<UserState>((set, get) => ({
```
theblockstalk, line 89:
```ts
if (e instanceof SdkError && e.code === SdkErrors.KeyNotFound) {
                await get().logout('Key not found on account');
                useErrorStore.getState().setError({ error: e, expected: false });
                set({ isAppInitialized: true });
            } else if (e instanceof SdkError && e.code === SdkErrors.AccountDoesntExist) {
                await get().logout('Account not found');
                set({ isAppInitialized: true });
            } else {
                throw e;
            }
```

src/store/userStore.ts
```diff
@@ -51,15 +51,10 @@ const useUserStore = create<UserState>((set, get) => ({
...
-
-        return status as UserStatus;
+        return get().status;
     },
     setStatus: async (newStatus: UserStatus) => {
-        const storeStatus = await AsyncStorage.setItem(STORAGE_NAMESPACE + 'store.status', newStatus);
+        await AsyncStorage.setItem(STORAGE_NAMESPACE + 'store.status', newStatus);
 
         set({ status: newStatus });
     },
```
theblockstalk, line 60:
```
logout: async (reason: string) => {
        await get().user.logout();
        if (get().status === UserStatus.LOGGED_IN) get().setStatus(UserStatus.NOT_LOGGED_IN);
        await printStorage('logout(): ' + reason);
    },
    ```

File: src/store/userStore.ts [hunk 6]
```diff
@@ -94,7 +89,7 @@ const useUserStore = create<UserState>((set, get) => ({
 
                 debug('network error condition status', status);
```
theblockstalk, line 91:
```ts
{
            debug('initializeStatusFromStorage() try');
            await get().user.initializeFromStorage();
            // get().setStatus(UserStatus.LOGGED_IN); // REDUNDANT: DELETE ME
            set({ isAppInitialized: true });
```

src/containers/MainSplashContainer.tsx
theblockstalk: lets take it back to like this https://github.com/Tonomy-Foundation/Tonomy-ID/compare/0.112.0...development#diff-f370f262227d9e004c0126fa7a26c164be8c8bd95a690ae2144e910267fa3f8bR31

src/containers/MainSplashContainer.tsx
```diff
@@ -54,16 +57,21 @@ export default function MainSplashScreenContainer({ navigation }: { navigation:
...
-                        //     } catch (e) {
-                        //         errorStore.setError({
-                        //             error: new Error('Error initializing wallet. Check your internet connection.'),
-                        //             expected: false,
-                        //         });
-                        //     }
-                        // }
+                        if (!initialized && isConnected) {
+                            try {
+                                await initializeWalletState();
```
theblockstalk, line 62: check network inside initializeWalletState() and if no network throw "Network request failed" error. Then you can use the more consistent logic here:
`progressiveRetryOnNetworkError(async () => await initializeWalletState());`

src/containers/MainSplashContainer.tsx
```diff
@@ -54,16 +57,21 @@ export default function MainSplashScreenContainer({ navigation }: { navigation:
...
-                        //     try {
-                        //         progressiveRetryOnNetworkError(async () => await initializeWalletState());
-                        //     } catch (e) {
-                        //         errorStore.setError({
-                        //             error: new Error('Error initializing wallet. Check your internet connection.'),
-                        //             expected: false,
-                        //         });
-                        //     }
-                        // }
+                        if (!initialized && isConnected) {
```
theblockstalk, line 60: we should delete this. `initializeWalletState()` will otherwise be called twice: once here and also again in the UserHome screen.

src/store/userStore.ts
```diff
@@ -81,14 +88,18 @@ const useUserStore = create<UserState>((set, get) => ({
...
+            debug('initializeStatusFromStorage() catch', e, typeof e);
+
             if (e instanceof SdkError && e.code === SdkErrors.KeyNotFound) {
                 await get().logout('Key not found on account');
+                set({ isAppInitialized: true });
                 useErrorStore.getState().setError({ error: e, expected: false });
             } else if (e instanceof SdkError && e.code === SdkErrors.AccountDoesntExist) {
                 await get().logout('Account not found');
+                set({ isAppInitialized: true });
             } else if (e.message === 'Network request failed') {
```
theblockstalk, line 103: looks like this `(e.message === 'Network request failed')` condition still needs to be deleted. See https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1162/files#r1765010945

src/store/useWalletStore.ts
```diff
@@ -66,36 +73,36 @@ const useWalletStore = create<WalletState>((set, get) => ({
...
-                if (
-                    (e.message && e.message === 'Network request failed') ||
-                    (e.msg && e.msg.includes('No internet connection'))
-                ) {
-                    throw new Error('Network request failed');
-                } else {
-                    debug('error when initializing wallet', e);
-                    throw new Error('Error initializing wallet, Check your internet connection');
-                }
+                console.error('error when initializing wallet', e);
```
theblockstalk, line 105: this also needs to `throw e;` right?

src/store/useWalletStore.ts
```diff
@@ -109,6 +116,8 @@ const useWalletStore = create<WalletState>((set, get) => ({
                 return;
             }
 
+            await connect();
```
theblockstalk, line 119: isn't this called in the MainContainer.tsx?

src/containers/MainContainer.tsx
```diff
@@ -81,21 +81,19 @@ export default function MainContainer({
...
+        if (!isConnected) {
+            debug('No network connection');
+        }
+    }, [isConnected]);
 
-        initializeAndFetchBalances();
-    }, [initializeWalletState, initialized, isConnected, errorStore]);
+    useEffect(() => {
+        if (!initialized) {
+            progressiveRetryOnNetworkError(async () => await initializeWalletState()).catch((e) => {
```
theblockstalk, line 91: this shouldn't be here, as it is called on the UserHome screen. Same reason as https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1162#discussion_r1765099269
Is that correct?

src/store/userStore.ts
theblockstalk: Sadia says if this condition is removed, then this shows error. Network error is not caught by progressiveRetryOnError() https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1162/files#r1765544429

https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1162/files#diff-7f3cb8966d073ff52698690db20e41f8606f8331b05bbd1bc116d447bdfa8e02R40 is not executing

PR #1182, opened 2024-10-01
Feature/listen esr protocol

sadiabbasi:


src/utils/chain/types.ts
```diff
@@ -367,4 +368,72 @@ export interface IChainSession {
...
+     * @param event The session event object.
+     */
+    onSessionEvent(event: unknown): void;
+
+    /**
+     * Handles deep linking
+     * @param url The deep link URL.
+     * @returns A boolean indicating if the transaction was successfully handled.
+     */
+    handleDeepLinking(url: string): Promise<boolean>;
```
theblockstalk, line 431: Similar to onTransactionEvent(), These will need to return some objects Which the CommunicationService can then use to determine if and where to send the user, and send those objects to the UI screen for use

src/utils/chain/types.ts
```diff
@@ -367,4 +368,72 @@ export interface IChainSession {
...
+    type?: string;
+};
+
+export interface IRequestListener extends IChainSession {
+    /**
+     * Processes incoming transaction events
+     * @param request The transaction request object.
+     * @returns A boolean indicating if the transaction was successfully handled.
+     */
+    onTransactionEvent(request: ESRTransactionRequest | EthereumTransactionRequest): Promise<{
```
theblockstalk, line 412: shouldn't this be `onTransactionEvent(request:ESRTransactionRequest | EthereumTransactionRequest)` ?

This is the original input that we get.

src/utils/chain/types.ts
```diff
@@ -367,4 +368,72 @@ export interface IChainSession {
...
+     * @returns A boolean indicating if the transaction was successfully handled.
+     */
+    handleDeepLinking(url: string): Promise<boolean>;
+
+    /**
+     * Handles QR scanning events and parsing.
+     * @param scannedData The scanned QR code data.
+     *  @returns A boolean indicating if the transaction was successfully handled.
+     */
+    handleQrScanning(scannedData: string): Promise<boolean>;
```
theblockstalk, line 438: Returning a boolen here is not enough. We need to tell the component what type of request that is, and other information so the component can decide what to do with the request, it also needs return values so that it can pass those to the next UI can show the correct interface to the user and process the request if the user approves.

src/utils/chain/types.ts
```diff
@@ -368,3 +369,41 @@ export interface IChainSession {
...
+    loginApp: ILoginApp;
+    privateKey: IPrivateKey;
+    account: IAccount;
+    request?: unknown;
+    reject(): Promise<void>;
+    approve(): Promise<void>;
+}
+
+export interface ITransactionRequest {
+    transactionRequest: ITransactionRequest;
```
theblockstalk, line 390: no need to have the object inside itself

src/utils/chain/types.ts
```diff
@@ -368,3 +369,41 @@ export interface IChainSession {
     rejectTransactionRequest(request: unknown): Promise<void>;
     getActiveAccounts(): Promise<IAccount[]>;
 }
+
+export interface ILoginApp {
+    getLogoUrl(): string;
```
theblockstalk, line 374: think - what app information do we need to display on the login screen
getOrigin()

src/utils/chain/types.ts
theblockstalk, line 1: do we also need to update ITransactionReceipt?

src/utils/chain/types.ts
```diff
@@ -368,3 +369,41 @@ export interface IChainSession {
...
+    privateKey: IPrivateKey;
+    account: IAccount;
+    request?: unknown;
+    reject(): Promise<void>;
+    approve(): Promise<ITransactionReceipt>;
+}
+
+export interface ISession {
+    initialize(): Promise<void>;
+    onQrScan(data: string): Promise<void>;
```
theblockstalk, line 402: onQrScan() and onLink() are both static functions. Can we express that in an interface?

prepare.app.config.ts
```diff
@@ -68,6 +68,12 @@ const expo: ExpoConfig = {
         bundleIdentifier: identifier,
         infoPlist: {
             NSCameraUsageDescription: 'We need access to your camera to scan the QR code.',
+            LSApplicationQueriesSchemes: ['esr'],
```
theblockstalk, line 71: should this also be registering "did" and the walletconnect schema?

src/utils/chain/types.ts
```diff
@@ -368,3 +369,58 @@ export interface IChainSession {
...
+    reject(): Promise<void>;
+    approve(): Promise<ITransactionReceipt>;
+}
+
+export interface ISession {
+    initialize(): Promise<void>;
+    onQrScan(data: string): Promise<void>; //make this function static
+    onLink(data: string): Promise<void>; //make this function static
+    onEvent(request: unknown): Promise<void>;
+    handleLoginRequest(request: unknown): Promise<void>;
```
theblockstalk, line 405: `handle...()` and `navigate...()` functions are not part of `ISession`, instead they are protected members of the implementation. We can put it in the AbstractSession class as protected member functions, if we want to declare it somewhere.

src/utils/chain/types.ts
```diff
@@ -368,3 +369,58 @@ export interface IChainSession {
...
+    constructor(chain: IChain) {
+        this.chain = chain;
+    }
+    async initialize(): Promise<void> {
+        // Implementation goes here
+    }
+    abstract onQrScan(data: string): Promise<void>;
+    abstract onLink(data: string): Promise<void>;
+    abstract onEvent(request: unknown): Promise<void>;
+    abstract handleLoginRequest(request: unknown): Promise<void>;
```
theblockstalk, line 422: should be protected for `handle` and `navigate` functions

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,54 @@
...
+import Web3Wallet from '@walletconnect/web3wallet';
+import NetInfo from '@react-native-community/netinfo';
+import Debug from 'debug';
+import { NETWORK_ERROR_MESSAGE } from '../../utils/errors';
+import useSessionStore from '../../store/sessionStore';
+import settings from '../../settings';
+
+const debug = Debug('tonomy-id:utils:initializeSession');
+
+export async function WaalletConnectSession() {
```
theblockstalk, line 11: `implements AbstractSession`

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,54 @@
...
+    const { protocols } = useSessionStore();
```
theblockstalk, line 12: you can only use a react hook (store) inside a component or another hook. this will not work.

Instead you need to do the reverse. The WalletConnectSession Will contain the logic to initialise the objects (core, web3WalletInstance) and store this inside its object. The WalletConnectSession object will then be saved in the useSessionStore()

src/utils/session/walletConnect.ts
theblockstalk: "WalletConnectSession"

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,54 @@
+import { Core } from '@walletconnect/core';
+import Web3Wallet from '@walletconnect/web3wallet';
+import NetInfo from '@react-native-community/netinfo';
+import Debug from 'debug';
+import { NETWORK_ERROR_MESSAGE } from '../../utils/errors';
+import useSessionStore from '../../store/sessionStore';
+import settings from '../../settings';
+
+const debug = Debug('tonomy-id:utils:initializeSession');
```
theblockstalk, line 9: `Debug('tonomy-id:utils:session:walletConnect');`

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,84 @@
...
+            } catch (e) {
+                console.error('useWalletStore() error on Web3Wallet.init()', JSON.stringify(e, null, 2));
+                if (e.msg && e.msg.includes('No internet connection')) throw new Error(NETWORK_ERROR_MESSAGE);
+                else throw e;
+            }
+        }
+    }
+
+    async onQrScan(data: string): Promise<void> {
+        debug(`QR code scanned with data: ${data}`);
```
theblockstalk, line 51: debug(`onQrScan() data ${data}`);
this is more consistent with the rest of the codebase. Also helps find the function easier. Please use this for `debug`

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,84 @@
...
+                        name: settings.config.appName,
+                        description: settings.config.ecosystemName,
+                        url: 'https://walletconnect.com/',
+                        icons: [settings.config.images.logo48],
+                    },
+                });
+
+                this.initialized = true;
+            } catch (e) {
+                console.error('useWalletStore() error on Web3Wallet.init()', JSON.stringify(e, null, 2));
```
theblockstalk, line 43: update string to match new class/function

src/store/sessionStore.ts
theblockstalk, line 1: yes this looks right

src/utils/session/walletConnect.ts
theblockstalk, line 1: need to finish the implementation now. removing the functionality you move here (e.g. the event listeners) from the rest of the codebase...

src/services/CommunicationModule.ts
```diff
@@ -32,6 +32,8 @@ import { SessionTypes, SignClientTypes } from '@walletconnect/types';
 import Debug from 'debug';
 import { isNetworkError, NETWORK_ERROR_MESSAGE } from '../utils/errors';
 import { debounce, progressiveRetryOnNetworkError } from '../utils/network';
+import { useSessionStore } from '../store/sessionStore';
+import { WalletConnectSession as WalletSession } from '../utils/session/walletConnect';
```
theblockstalk, line 36: `import { WalletConnectSession } from '../utils/session/walletConnect';`

src/store/sessionStore.ts
```diff
@@ -0,0 +1,12 @@
...
+import { WalletConnectSession } from '../utils/session/walletConnect';
+
+interface SessionState {
+    walletConnectSession: WalletConnectSession | null;
+    setWalletConnectSession: (session: WalletConnectSession) => void;
+}
+
+export const useSessionStore = create<SessionState>((set) => ({
+    walletConnectSession: null,
+    setWalletConnectSession: (session: WalletConnectSession) => set({ walletConnectSession: session }),
```
theblockstalk, line 11: instead of setter for each session, we can have 1x initialize function here that creates all the session objects and inializeses them. Will make it less messy in the components. What do you think?

sadiabbasi: will do when implement for antelope too..

src/services/CommunicationModule.ts
theblockstalk, line 1: very nice to see this logic taken out of here. Keep the UI component clean with a its own relevant logic only. That is abiding by the "single responsibility principal" well ⭐ ⭐ ⭐ ⭐ ⭐

src/services/CommunicationModule.ts
```diff
@@ -44,14 +30,37 @@ export default function CommunicationModule() {
...
+        if (walletConnectSession && walletConnectSession.initialized) {
+            debug('initializeWalletState() Already initialized');
+            return;
         }
-    }, [initializeWalletState, initialized]);
+
+        try {
+            const session = new WalletConnectSession();
+
+            await session.initialize();
```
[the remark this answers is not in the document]
theblockstalk: ahhh. just seen your comment here https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1182/files#r1821536078
understood 👍

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+    findEthereumTokenByChainId,
+    getAccountFromChain,
+    getKeyFromChain,
+    getKeyOrNullFromChain,
+    tokenRegistry,
+} from '../tokenRegistry';
+
+const debug = Debug('tonomy-id:utils:session:walletConnect');
+
+export class WalletConnectLoginApp implements ILoginApp {
```
theblockstalk, line 44: this does not seem very WalletConnect specific. Perhaps this can be moved to `types` and called LoginApp

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+    loginApp: ILoginApp;
+    account: IAccount[];
+    request: SignClientTypes.EventArguments['session_proposal'];
+    session: ISession;
+    namespaces: SessionTypes.Namespaces;
+
+    constructor(
+        loginApp: ILoginApp,
+        accounts: IAccount[],
+        request: SignClientTypes.EventArguments['session_proposal'],
```
theblockstalk, line 85: this might be optional? same with session?

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+    }
+    async reject(): Promise<void> {
+        await this.session.web3wallet?.rejectSession({
+            id: this.request.id,
+            reason: getSdkError('USER_REJECTED'),
+        });
+    }
+
+    async approve(): Promise<void> {
+        await this.session.web3wallet?.approveSession({
```
theblockstalk, line 113: will this also sign the transaction?

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+                id: this.request.id,
+                error: getSdkError('USER_REJECTED'),
+                jsonrpc: '2.0',
+            };
+
+            await this.session.web3wallet?.respondSessionRequest({
+                topic: this.request.topic,
+                response,
+            });
+        } else throw new Error('Session or request is not defined');
```
theblockstalk, line 162: if there is no request and session we need to handle this. it may be an asset transfer from the AssetsDetailsContainer. Same with `approve()`

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+        } else {
+            // Step 2: Check for any missing keys
+            for (const chainId of chainIds) {
+                const chainEntry = findEthereumTokenByChainId(chainId);
+
+                if (chainEntry) {
+                    const key = await getKeyOrNullFromChain(chainEntry);
+
+                    if (!key) {
+                        // navigation.navigate('CreateEthereumKey', {
```
theblockstalk, line 280: should be uncommented? should use `navigateToEthereumKey()` function

src/utils/session/walletConnect.ts
theblockstalk, line 1: see refactor branch that I made. This will need to be updated based on the centralized logic it uses...

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+                return;
+            }
+        }
+    }
+
+    protected async navigateToLoginScreen(request: ILoginRequest): Promise<void> {
+        navigate('WalletConnectLogin', { loginRequest: request });
+    }
+
+    protected async navigateToTransactionScreen(request: ITransactionRequest): Promise<void> {
```
theblockstalk, line 397: use the `WalletConnectTransactionRequest` type here, which will give better type safety while still complying to `ITransactionRequest`. Do the same for `navigateToLoginScreen(request: ILoginRequest)`

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+    EthereumAccount,
+    EthereumChain,
+    EthereumPrivateKey,
+    EthereumToken,
+    EthereumTransaction,
+    EthereumTransactionReceipt,
+} from '../chain/etherum';
+import { navigate } from '../../services/NavigationService';
+import {
+    eip155StringToChainId,
```
theblockstalk, line 34: move this and `findEthereumTokenByChainId()` to this file as this is more relevant place

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,407 @@
...
+
+export class WalletTransactionRequest implements ITransactionRequest {
+    transaction: ITransaction;
+    privateKey: IPrivateKey;
+    account: IAccount;
+    request?: SignClientTypes.EventArguments['session_request'];
+    session?: ISession;
+    origin: string;
+
+    constructor(
```
theblockstalk, line 128: you could use an alternative constructure which also does some of the work currently in the `handleTransactionRequest()` function which is relevant just for this object. It will be a bit cleaner

This question is - where else will these object be created and what will be the same information that are available at that time. I think that can happen during the `handleTransactionRequest()` but also from the `SendAssetsContainer()`. Data that is available in these both these places should be part of the alternative constructor, while if available in only one place should not.

You can also do the same for the `WalletConnectLoginRequest` by moving relevant logic to create this object into an alternative constructor.

src/services/CommunicationModule.ts
```diff
@@ -169,6 +178,7 @@ export default function CommunicationModule() {
 
     useEffect(() => {
         progressiveRetryOnNetworkError(loginToService);
+        progressiveRetryOnNetworkError(initializeWalletConnect);
```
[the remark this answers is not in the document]
sadiabbasi: mistakenly added i will remove one...

src/utils/session/walletConnect.ts
sadiabbasi: for login request is required parameter and i guess same for session

src/utils/session/walletConnect.ts
sadiabbasi: for sign transaction it uses web3wallet.respondSessionRequest function

src/containers/CreateEthereumKeyContainer.tsx
```diff
@@ -118,32 +122,16 @@ export default function CreateEthereumKeyContainer({
...
-                    if (!chainEntry) throw new Error('Chain not found');
-                    const privateKey = await getKeyFromChain(chainEntry);
-
-                    navigation.navigate('SignTransaction', {
-                        transaction,
-                        privateKey,
-                        request: route.params.payload as Web3WalletTypes.SessionRequest,
-                        session: walletsession,
-                        origin: verifyContext?.verified?.origin,
-                    });
```
sadiabbasi, line 146: can you check this i removed all the re-implemented logic which we already have in handleTransactionRequest... so i just call the handleTransactionRequest and remove this logic here.. let me know your thoughts on this

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -254,8 +221,8 @@ export default function SignTransactionConsentContainer({
 
             navigation.navigate('Assets');
 
-            if (session) {
-                await session.rejectTransactionRequest(request);
+            if (request.session) {
```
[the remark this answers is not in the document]
sadiabbasi: fixed....

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,442 @@
...
+    account: IAccount;
+    request?: SignClientTypes.EventArguments['session_request'];
+    session?: ISession;
+    origin: string | null;
+
+    constructor(
+        transaction: ITransaction,
+        privateKey: IPrivateKey,
+        account: IAccount,
+        session?: ISession,
```
theblockstalk, line 123: good alternative constructors. However I dont think it is a good idea to use optional params in the normal constructor. Better to make some private setters that can be used in the alternative constructors to finish making the object. aka `private setSession()` and `private setRequest()` which are used in `fromRequest()`

src/utils/session/walletConnect.ts
```diff
@@ -0,0 +1,442 @@
...
+        transaction?: ITransaction
+    ): Promise<void> {
+        const params = {
+            requestType,
+            request,
+            session,
+        };
+
+        if (transaction) {
+            params['transaction'] = transaction;
```
theblockstalk, line 437: better use `params.transaction` for type safety. probably need to declare the params type on L430

src/containers/CreateEthereumKeyContainer.tsx
theblockstalk: if it is working then it looks fine to me. i think it makes sense that this can work.

PR #1214, opened 2024-10-21
Feature/vne wallet UI jekin

jekingohel:


src/utils/StorageManager/repositories/assetStorageManager.ts
```diff
@@ -34,10 +34,13 @@ export abstract class AssetStorageManager {
...
-
-                if (usdBalance > 0 && usdBalance !== existingAsset.usdBalance) {
-                    existingAsset.usdBalance = usdBalance;
+                if (asset.getToken().getSymbol() !== 'SepoliaETH') {
+                    const usdBalance = await asset.getUsdValue();
+                    if (usdBalance > 0 && usdBalance !== existingAsset.usdBalance) {
+                        existingAsset.usdBalance = usdBalance;
+                    }
+                } else {
+                    existingAsset.usdBalance = 0;
```
sadiabbasi, line 43: don't need the check here. update it in the Chain level

src/context/AppSettingsContext.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { createContext, useState, useEffect, useContext } from 'react';
```
sadiabbasi, line 1: we don't need this context... and also revert where you are implemented

src/containers/SendAssetContainer.tsx
```diff
@@ -19,7 +19,16 @@ export type SendAssetProps = {
...
+export enum ApplicationError {
+    NotEnoughCoins = 'NotEnoughCoins',
+}
+export class AppError extends Error {
+    public readonly type: ApplicationError;
+    constructor(type: ApplicationError, message: string) {
+        super(message);
+        this.type = type;
+    }
+}
```
sadiabbasi, line 31: why you added this here?

src/containers/SendAssetContainer.tsx
```diff
@@ -75,12 +84,10 @@ const SendAssetContainer = (props: SendAssetProps) => {
...
-                errorStore.setError({
-                    error: new Error('You do not have enough balance to perform transaction!'),
-                    expected: true,
-                    title: 'Insufficient balance',
-                });
-                return;
+                throw new AppError(
+                    ApplicationError.NotEnoughCoins,
+                    'You do not have enough balance to perform this transaction!'
+                );
```
sadiabbasi, line 90: use this NotEnoughCoins from ApplicationError which we already have inside utils/errors.ts

src/utils/StorageManager/repositories/assetStorageManager.ts
```diff
@@ -35,8 +35,7 @@ export abstract class AssetStorageManager {
 
             try {
                 const usdBalance = await asset.getUsdValue();
-
-                if (usdBalance >= 0 && usdBalance !== existingAsset.usdBalance) {
+                if (usdBalance > 0 && usdBalance !== existingAsset.usdBalance) {
```
sadiabbasi, line 38: >= condition works for sepolia otherwise it will not override the existing user usd values for sepolia

src/utils/chain/types.ts
```diff
@@ -173,6 +173,9 @@ export abstract class AbstractAsset implements IAsset {
     }
 
     async getUsdValue(): Promise<number> {
+        if (this.token.getSymbol() === 'SepoliaETH') {
+            return 0;
+        }
```
sadiabbasi, line 178: don't add this here. i added this inside the EthereumToken getUsdPrice()

Issue #1220, opened 2024-10-23
console.error is causing "Unexpected JS" errors to be thrown


theblockstalk:
caused in `/home/dev/Documents/Git/Tonomy/Tonomy-ID/src/components/TErrorModal.tsx`

```log
 DEBUG  tonomy-id:utils:exceptions Unexpected JS Error Logs +43s TErrorModal() unexpected error string "TErrorModal() unexpected error"
 DEBUG  tonomy-id:store:errorStore setError +194ms TErrorModal() unexpected error
```

theblockstalk:
this can be tested by calling `console.error` or by calling `TErrorModal` with `expected: true` which will result in recursive rendering

theblockstalk:
lets merge sentry PR, and if we see this again we should then action

theblockstalk:
this is no longer observed

PR #1221, opened 2024-10-23
fix: fixed signing error message, logging and error handling

theblockstalk:


App.tsx
```diff
@@ -1,6 +1,5 @@
-// IMPORTANT: The following 3 packages should be imported in this order:
+// IMPORTANT: The following 2 packages should be imported in this order:
 import 'reflect-metadata';
-import './src/utils/logs';
```
theblockstalk, line 3: someting in this file was causing https://github.com/Tonomy-Foundation/Tonomy-ID/issues/1220. Removing the file solved the issue. The `log.ts` service was a nice to have but not worth having if we could not solve this error (which I could not)

src/components/TErrorModal.tsx
```diff
@@ -101,7 +99,7 @@ export default function TErrorModal(props: TErrorModalProps) {
                 title={props.title ?? 'Something went wrong'}
                 iconColor={theme.colors.error}
             >
-                {/* {props.children} */}
+                {props.children}
```
sadiabbasi, line 102: children is an optional parameter in the props... will it work if there is no props.children?

src/utils/chain/antelope.ts
```diff
@@ -180,14 +181,16 @@ export class AntelopePrivateKey extends AbstractPrivateKey implements IPrivateKe
...
-
         try {
+            const transaction = await this.signTransaction(data);
+
             const receipt = await this.chain.getApiClient().v1.chain.push_transaction(transaction);
 
             return new AntelopeTransactionReceipt(this.chain, receipt, transaction);
         } catch (error) {
-            console.error('sendTransaction()', JSON.stringify(error, null, 2));
+            if (error?.message?.includes('Provided keys, permissions, and delays do not satisfy declared authorizations at')) {
```
sadiabbasi, line 191: didn't this throw any code.. that will looks cleaner..

src/components/TErrorModal.tsx
theblockstalk: yes it seems to work if there are no children

PR #1235, opened 2024-11-01
Updated Expo SDK to 51

theblockstalk:
- upgraded expo from v49 to v51
  - fixed `expo-camera`
  - using the `expo-sqlite/legacy` export to keep compatibility with `typeorm`
- updated all other packages (exceptions: https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1235#issuecomment-2459709467)
  - fixed `react-native-paper` type errors and theme by keeping the old MD2 theme
  - fixed lint and type errors
- moved `eas-cli` to be global install as per expo recommendation (this is to avoid package version bugs)
- added instructions in readme to ensure expo packages are right version
- removed invaluable screens and moved unused containers to `./src/containers/unused`


theblockstalk:
all packages upgradedexcept:
- new eslint v9 which requires us to change to new format. Should probably switch to use the expo lint support https://docs.expo.dev/guides/using-eslint/
- latest version of @wharfkit/antelope (needs to be updated in SDK first)

sadiabbasi (approved):

PR #1246, opened 2024-11-07
Feature/1238 antelope i session

sadiabbasi:


src/utils/session/antelope.ts
```diff
@@ -0,0 +1,264 @@
...
+                });
+
+                debug('approveTransactionRequest() response status', response.status);
+
+                if (!response.ok || response.status !== 200) {
+                    console.error(`Failed to send callback: ${JSON.stringify(response)}`);
+                }
+            }
+        } else {
+            receipt = await this.privateKey.sendTransaction(this.transaction as AntelopeTransaction);
```
theblockstalk, line 144: move this outside the condition and remove the condition

theblockstalk: also should handle teh case where `sendTransaction()` throws an error and there is a session and callback - in which case you should also send a reject to the callback

src/utils/session/antelope.ts
```diff
@@ -0,0 +1,264 @@
...
+        );
+
+        this.navigateToTransactionScreen(transactionRequest);
+    }
+
+    protected async navigateToLoginScreen(request: unknown): Promise<void> {
+        //TODO when implement handle identity request
+    }
+
+    protected async navigateToTransactionScreen(request: AntelopeTransactionRequest): Promise<void> {
```
theblockstalk, line 253: the `navigateTo...()` functions are probably the same for all ISession classes and can therefore be moved and implemented in `AbstractSession`

src/utils/session/walletConnect.ts
```diff
@@ -148,11 +144,13 @@ export class WalletTransactionRequest implements ITransactionRequest {
     static async fromRequest(
         chainEntry: TokenRegistryEntry,
         ethereumPrivateKey: EthereumPrivateKey,
-        request: SignClientTypes.EventArguments['session_request'],
+        event: SignClientTypes.EventArguments['session_request'],
```
theblockstalk, line 147: request?

src/config/config.testnet.json
```diff
@@ -13,7 +13,7 @@
         "lightBackground": "#F6F9FB"
     },
     "appLogoUrl": "./src/assets/pangea/favicon.png",
-    "appName": "Pangea Testnet",
+    "appName": "Tonomy ID Development",
```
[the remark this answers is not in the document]
sadiabbasi: i am using the VS code but this change is needed to test the antelope that's why i keep this once all finalized i will remove

src/utils/session/antelope.ts
```diff
@@ -0,0 +1,267 @@
...
+                        console.error(`Failed to send callback: ${JSON.stringify(response)}`);
+                    }
+                }
+            }
+
+            return receipt;
+        } catch (e) {
+            console.error('Error approving transaction', e);
+
+            if (this.resolvedSigningRequest) {
```
theblockstalk, line 156: you dont need this condition guard here, it exists inside `reject()`

theblockstalk (commented):
1 comment + this is not resolved https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1246#discussion_r1833204812

PR #1250, opened 2024-11-12
Feature/1226 signing edge cases

jekingohel:


src/components/NegligibleTransactionFees.tsx
```diff
@@ -0,0 +1,78 @@
...
+    onClose: () => void;
+    transactionFee: TransactionFeeData;
+};
+
+const NegligibleTransactionFees = (props: props) => {
+    return (
+        <RBSheet
+            ref={props.refMessage}
+            openDuration={150}
+            closeDuration={100}
```
sadiabbasi, line 19: what this duration will do?

src/components/NegligibleTransactionFees.tsx
```diff
@@ -0,0 +1,78 @@
...
+                <Text style={styles.drawerHead}>Transaction fee</Text>
+                <TouchableOpacity onPress={props.onClose}>
+                    <TIconButton icon={'close'} color={theme.colors.lightBg} iconColor={theme.colors.grey1} />
+                </TouchableOpacity>
+            </View>
+            <ScrollView>
+                <View style={styles.content}>
+                    <View>
+                        <Text style={{ ...commonStyles.primaryFontFamily, fontSize: 17 }}>
+                            {props.transactionFee.fee}
```
sadiabbasi, line 33: In figma it also display symbol i.e. 0.003453 LEOS

src/components/Transaction.tsx
```diff
@@ -106,33 +110,82 @@ export function ContractOperationDetails({ operation, date }: { operation: Opera
...
             </View>
-            <View style={{ flexDirection: 'row', justifyContent: 'space-between', marginTop: 10 }}>
+            <View style={{ flexDirection: 'row', justifyContent: 'space-between' }}>
                 <Text style={styles.secondaryColor}>Function:</Text>
-                <Text>{operation.functionName}</Text>
+                {operation.functionName ? (
+                    <Text style={{ color: theme.colors.secondary }}>{`${operation.functionName}()`}</Text>
+                ) : (
+                    <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+                        <TouchableOpacity onPress={() => setFunToolTipVisible(true)}>
```
sadiabbasi, line 119: unknown word is not clickable i think the icon is clickable.. Check with pavel too.

src/components/Transaction.tsx
```diff
@@ -106,33 +110,82 @@ export function ContractOperationDetails({ operation, date }: { operation: Opera
...
+                            <IconButton
+                                style={{ height: 20 }}
+                                icon={Platform.OS === 'android' ? 'chevron-up' : 'chevron-up'}
+                                size={Platform.OS === 'android' ? 18 : 22}
+                            />
+                        )}
+                    </TouchableOpacity>
+                ) : (
+                    <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+                        <TouchableOpacity onPress={() => setTranToolTipVisible(true)}>
```
sadiabbasi, line 168: same as transaction fee

src/components/Transaction.tsx
```diff
@@ -174,16 +233,23 @@ export function TransactionFee({ transactionFee }: { transactionFee: Transaction
...
-                    <Text style={[styles.secondaryColor, commonStyles.secondaryFontFamily]}>
-                        (${transactionFee.usdFee})
-                    </Text>
+                    <Text>{transactionFee.feeLabel ?? transactionFee.fee}</Text>
+                    {transactionFee.feeLabel === null && (
+                        <Text style={[styles.secondaryColor, commonStyles.secondaryFontFamily]}>
+                            (${transactionFee.usdFee})
+                        </Text>
+                    )}
+                    {transactionFee.feeLabel === 'negligible' && (
```
sadiabbasi, line 248: we can calculate this as if the usd fee is less than 0.001 we can add 'negligible' no need for feeLabel property

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -31,6 +34,9 @@ export default function SignTransactionConsentContainer({
 }) {
     const { transaction } = request;
 
+    const [time, setTime] = useState(120);
+    const [expired, setExpired] = useState(false);
```
sadiabbasi, line 38: add the expiration time in the abstraction level as it's only for antelope. create somefunction inside ITransaction

src/components/Transaction.tsx
theblockstalk: better make both clickable as the icon on the screen is so small. makes it easier for the user if both are clickable and that is more intuitive

src/containers/SignTransactionConsentContainer.tsx
theblockstalk: ITransaction.getExpiration(): Date | null

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -31,6 +34,9 @@ export default function SignTransactionConsentContainer({
 }) {
     const { transaction } = request;
 
+    const [time, setTime] = useState(120);
```
theblockstalk, line 37: `useState(null)` make it that transactions do not expire by default as this is only part of the Antelope protocol

src/components/NegligibleTransactionFees.tsx
jekingohel: openDuration: Duration of the animation when opening bottom sheet.
closeDuration: Duration of the animation when closing bottom sheet.

src/components/NegligibleTransactionFees.tsx
jekingohel: transactionFee.fee contain symbol, so its not require to add symbol

src/components/Transaction.tsx
```diff
@@ -157,33 +195,54 @@ export function ContractOperationDetails({ operation, date }: { operation: Opera
 
 export function TransactionFee({ transactionFee }: { transactionFee: TransactionFeeData }) {
     const [toolTipVisible, setToolTipVisible] = useState(false);
+    const usdFeeFloat = parseFloat(transactionFee.usdFee);
+    const refMessage = useRef<{ open: () => void; close: () => void }>(null);
+    const isNegligible = usdFeeFloat <= 0.001 && usdFeeFloat > 0;
+    const isFree = usdFeeFloat === 0;
```
sadiabbasi, line 201: isFree is always 0 ?

src/components/Transaction.tsx
```diff
@@ -157,33 +195,54 @@ export function ContractOperationDetails({ operation, date }: { operation: Opera
...
+                        <TouchableOpacity onPress={() => refMessage.current?.open()} style={{ marginLeft: 2 }}>
+                            <Text>negligible</Text>
+                            <WarningCircle width={15} height={15} color={theme.colors.success} />
+                        </TouchableOpacity>
+                    )}
+                    {!isNegligible && (
+                        <Text style={[styles.secondaryColor, commonStyles.secondaryFontFamily]}>
+                            (${transactionFee.usdFee})
+                        </Text>
+                    )}
```
sadiabbasi, line 245: we can simplify condition like {isNeglibible? ( <TouchableOpacity ...... /> ) : <Text....>}

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -106,17 +114,26 @@ export default function SignTransactionConsentContainer({
     }, [transaction, getOperationData]);
 
     const fetchTransactionFee = useCallback(async () => {
-        const fee = await transaction.estimateTransactionFee();
+        const fee = await request.transaction.estimateTransactionFee();
```
sadiabbasi, line 117: why change this? as we added     const { transaction } = request;

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -218,51 +240,76 @@ export default function SignTransactionConsentContainer({
...
+        );
+    };
+
+    useEffect(() => {
+        const intervalId = setInterval(() => {
+            const expiration =
+                request.transaction.getExpiration() && request.account ? request.transaction.getExpiration() : null;
+            if (!expiration) {
+                clearInterval(intervalId);
+                return;
```
sadiabbasi, line 279: i think add this condition before the setInterval

src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -218,51 +240,76 @@ export default function SignTransactionConsentContainer({
...
+                const minutes = Math.floor(timeDiff / 1000 / 60);
+                const seconds = Math.floor((timeDiff / 1000) % 60);
+                setRemainingTime(`${String(minutes).padStart(2, '0')}:${String(seconds).padStart(2, '0')}`);
+            }
+        }, 1000);
+        return () => clearInterval(intervalId);
+    }, [request]);
+
+    const renderExpirationTimer = () => {
+        return !expired && request.account && transaction.getExpiration() ? (
```
sadiabbasi, line 298: we can simplify the condition here too.. right @theblockstalk

theblockstalk: possible but not required

theblockstalk (approved):

src/components/Transaction.tsx
jekingohel: it wont be always 0 because it's a conditional value so it will be boolean type

src/containers/SignTransactionConsentContainer.tsx
jekingohel: Because its require to get request.account to check its assets transaction or not.

PR #1255, opened 2024-11-14
Feature/1240 antelope login request

sadiabbasi:


src/utils/session/antelope.ts
```diff
@@ -246,12 +382,12 @@ export class AntelopeSession extends AbstractSession {
...
             data: action.data,
         }));
-        const account = AntelopeAccount.fromAccount(this.chain, accountName);
+        const account = AntelopeAccount.fromAccount(this.chain, this.accountName);
         const transaction = AntelopeTransaction.fromActions(actions, this.chain, account);
 
         const callback = resolvedSigningRequest.request.data.callback;
 
-        new LoginApp('', callback, '', [this.chain]);
+        new LoginApp(this.chain.getName(), callback, this.chain.getLogoUrl(), [this.chain]);
```
theblockstalk, line 390: why do we need this object? seems unused

src/utils/session/antelope.ts
```diff
@@ -163,9 +279,12 @@ export class AntelopeSession extends AbstractSession {
     privateKey: AntelopePrivateKey;
```
theblockstalk, line 279: does this need privateKey still? both the Login and Transaction request should have it but I don't think this is needed at the session level anymore.

Issue #1178, opened 2024-09-30
Hcaptcha stuck on black screen when turn off internet in verify hcaptcha screen


sadiabbasi:
Acceptance criteria:

STeps to reproduce

- [ ] create account with turn on internet.
- [ ] and when come to hcaptcha screen  and turn off internet it shows black screen and stcuk there.
- [ ] instead of stuck on that screen throw some error

sadiabbasi:
fixed this in PR https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1255

PR #1273, opened 2024-11-28
Bug fixes: #1178, 291, 1271

sadiabbasi:


src/containers/AssetsContainer.tsx
```diff
@@ -119,10 +118,11 @@ export default function AssetsContainer({ navigation }: { navigation: AssetsScre
         }
     }, [updateAllBalances]);
 
-    // updateAllBalances() on mount and every 20 seconds
     useEffect(() => {
-        updateAllBalances();
+        fetchCryptoAssets().then(() => setAssetLoading(false));
```
theblockstalk, line 122: this will cause fetchCryptoAssets to be called on render, but not to update balances... why is this better...

sadiabbasi: this will call fetchCryptoAssets  on render ... and then also call after updateBalance in line 100..

before it fetchCryptoAssets call after updateBalance on render and first time user open screen update balance throw error because first render fetchAsset call after updateBalance..

src/containers/AssetsContainer.tsx
```diff
@@ -38,7 +38,11 @@ export default function AssetsContainer({ navigation }: { navigation: AssetsScre
 
     const fetchCryptoAssets = useCallback(async () => {
         try {
-            if (!accountsInitialized) await initializeWalletAccount(user);
+            if (!accountsInitialized) {
```
theblockstalk, line 41: - move this check into `updateAllBalances()`, before `updateBalance()` is called
- remove `updateBalance()` from this if condition

package.json
```diff
@@ -38,7 +38,7 @@
         "@react-navigation/native": "^6.1.18",
         "@react-navigation/native-stack": "^6.11.0",
         "@sinonjs/text-encoding": "^0.7.3",
-        "@tonomy/tonomy-id-sdk": "0.24.0",
+        "@tonomy/tonomy-id-sdk": "0.24.0-rc.1",
```
theblockstalk, line 41: change to latest version 0.24. the `-rc.` is a previous version.

theblockstalk (changes requested):

PR #1296, opened 2024-12-09
Feature/1193 withdraw success UI

sadiabbasi:


src/containers/AssetsContainer.tsx
```diff
@@ -194,10 +194,17 @@ export default function AssetsContainer({ navigation }: { navigation: AssetsScre
                                 <TouchableOpacity
                                     key={index}
                                     onPress={() => {
-                                        navigation.navigate('AssetDetail', {
-                                            screenTitle: chainName,
-                                            chain: chainObj.chain,
-                                        });
+                                        if (chainObj.token.getSymbol() === 'LEOS') {
```
theblockstalk, line 197: why create a new screen for this? this design is meant to apply for all assets

src/containers/SuccessLeosWithDrawContainer.tsx
theblockstalk, line 1: this should be VestedAssetsWithdrawSuccessContainer.tsx (abstracted, not necessarily for LEOS)

src/utils/chain/antelope.ts
```diff
@@ -453,7 +453,7 @@ export const TonomyLocalChain = new AntelopeChain(
     'Tonomy Localhost',
     'unknown chain id at this time',
     'https://github.com/Tonomy-Foundation/documentation/blob/master/images/logos/Pangea%20256x256.png?raw=true',
-    'https://local.bloks.io/?nodeUrl=https%3A%2F%2Fblockchain-api-staging.tonomy.foundation&coreSymbol=LEOS&corePrecision=6&systemDomain=eosio',
+    'https://local.bloks.io',
```
theblockstalk, line 456: this should not be necessary.
the chain.getExplorerUrl() should be adding all those details already in the respective abstractions

src/containers/AssetsContainer.tsx
sadiabbasi: if that the case then will remove this condition

src/utils/chain/antelope.ts
sadiabbasi: if we keep this it again concetenate inside the getExplorerUrl and for staging and development block explorer url not working.

theblockstalk: should be fixed in the staging/development chain implementation then, not here

sadiabbasi:
created with draw success screen in different PR i.e  https://github.com/Tonomy-Foundation/Tonomy-ID/pull/1410 .. so this is not needed anymore..

PR #1313, opened 2024-12-13
Feature/fix todo signing edge cases

theblockstalk:


src/containers/SignTransactionConsentContainer.tsx
```diff
@@ -234,37 +233,59 @@ export default function SignTransactionConsentContainer({
...
-                </View>
-            );
+        function Logo() {
+            if (!origin) {
+                return <Image style={styles.logo} source={{ uri: chainLogo }} />;
+            } else {
+                try {
+                    return (
+                        <Image
+                            style={[styles.logo, commonStyles.marginBottom]}
```
theblockstalk, line 248: calling an external source. an http error will cause the whole component to fail so must catch the error

PR #1346, opened 2025-01-07
feat: :zap: changed parseFloat to decimal

sadiabbasi:


src/utils/chain/antelope.ts
```diff
@@ -620,7 +621,7 @@ function getAssetFromQuantity(quantity: string, chain: AntelopeChain): IAsset {
     const precision = amountString.split('.')[1].length;
 
     const token = new AntelopeToken(chain, name, symbol, precision, '', '');
-    const amountNumber = parseFloat(amountString);
+    const amountNumber = new Decimal(amountString).toNumber();
```
theblockstalk, line 624: this wont work, will still end up in floating point inaccuracy. we need to do BOTH the number conversion and the arithmetic using Decimal. to do this we would need to replace the Asset use of Decimal type instead of Bignumber for the `amount` which requires a bit more work.

src/utils/chain/types.ts
theblockstalk, line 1: same as above comment. need to do the arithmetic using Decimal, not just the conversion. otherwise it will not work.

src/utils/chain/types.ts
```diff
@@ -182,31 +183,31 @@ export interface IAsset {
...
-        return new Asset(this.token, this.amount + other.getAmount());
+        return new Asset(this.token, this.amount.add(other.getAmount()));
     }
     async getUsdValue(): Promise<number> {
         const price = await this.token.getUsdPrice();
 
         if (price) {
             const precision = this.token.getPrecision();
-            const precisionMultiplier = BigInt(10) ** BigInt(precision);
+            const precisionMultiplier = new Decimal(10).pow(precision);
```
theblockstalk, line 203: can siplify all this logic. turn `price` into a Decimal, then multiply

src/utils/chain/types.ts
```diff
@@ -218,14 +219,12 @@ export abstract class AbstractAsset implements IAsset {
         return this.token.getPrecision();
     }
     printValue(precision?: number): string {
-        const amountNumber = Number(this.amount);
-        const precisionNumber = Number(10 ** this.token.getPrecision());
+        const precisionNumber = new Decimal(10).pow(this.token.getPrecision());
```
theblockstalk, line 222: most of this logic not needed anymore. can just use Decimal.toString() but might need to truncate the string for the precision

src/utils/chain/antelope.ts
```diff
@@ -376,8 +377,8 @@ export class AntelopeToken extends AbstractToken implements IToken {
         );
         const asset = assets.find((asset) => asset.symbol.toString() === this.toAntelopeSymbol().toString());
 
-        if (!asset) return new Asset(this, BigInt(0));
-        return new Asset(this, BigInt(asset.units.value));
+        if (!asset) return new Asset(this, new Decimal(0));
+        return new Asset(this, new Decimal(asset.units.value));
```
theblockstalk, line 381: dont think this works

previously for 1 LEOS
 - precision = 4
 - amount = 10000n
 
what we need now is
- precision = 4
- amount = Decimal('1')

please check

src/utils/chain/antelope.ts
```diff
@@ -413,7 +414,7 @@ class PangeaVestedToken extends AntelopeToken {
 
         const vestedBalance = await vestingContract.getBalance(lookupAccount.getName());
 
-        return new Asset(this, BigInt(vestedBalance * 10 ** this.precision));
+        return new Asset(this, new Decimal(vestedBalance).mul(new Decimal(10).pow(this.precision)));
```
theblockstalk, line 417: dont need to multiply by precision all the time, can just use Decimal objects directly. please check

src/utils/chain/antelope.ts
```diff
@@ -676,29 +677,29 @@ export class AntelopeTransaction implements ITransaction {
...
-
         // const operationAmounts = (await Promise.all(operationAmountsPromises)).reduce((a, b) => a + b, BigInt(0));
         //     this return string
 
         // let amount = operationAmounts + (await this.estimateTransactionFee()).getAmount();
         // Ensure all operation amounts are BigInt
         const operationAmounts = (await Promise.all(operationAmountsPromises)).reduce((a, b) => {
-            return BigInt(a) + BigInt(b);
-        }, BigInt(0));
+            return new Decimal(a).plus(new Decimal(b));
```
theblockstalk, line 693: L685 already returns a Decimal, no need to create new ones

src/utils/chain/etherum.ts
```diff
@@ -195,7 +196,7 @@ export class EthereumToken extends AbstractToken {
 
         debug('getBalance() balance', balanceWei);
 
-        return new Asset(this, balanceWei);
+        return new Asset(this, new Decimal(balanceWei.toString()));
```
theblockstalk, line 199: 0.0001 ETH = 1000000000000000 Wei. so you want to create Decimal('0.001)') not Decimal('1000000000000000') which is what this code is doing

src/utils/chain/etherum.ts
```diff
@@ -372,10 +376,10 @@ export class EthereumTransaction implements ITransaction {
 
         const totalGasFee = feeData.gasPrice ? wei * feeData.gasPrice : wei;
 
-        return new Asset(this.chain.getNativeToken(), totalGasFee);
+        return new Asset(this.chain.getNativeToken(), new Decimal(totalGasFee.toString()));
```
theblockstalk, line 379: similar. need to check if this is the bigint version with precision, or the ETh value

theblockstalk: also L382, 426

src/utils/chain/etherum.ts
sadiabbasi: i tested this is fine because for display we use printValue or .toSTring() function and there we are fixing this...

src/utils/chain/etherum.ts
sadiabbasi: 1|id  |  LOG  totalGasFee 408492287784000n

src/utils/chain/etherum.ts
```diff
@@ -372,10 +376,10 @@ export class EthereumTransaction implements ITransaction {
...
     }
     async estimateTransactionTotal(): Promise<Asset> {
-        const amount = (await this.getValue()).getAmount() + (await this.estimateTransactionFee()).getAmount();
+        const amount = (await this.getValue()).getAmount().plus((await this.estimateTransactionFee()).getAmount());
```
sadiabbasi, line 382: amount +5s "476547762661000"

package.json
```diff
@@ -55,6 +55,7 @@
         "bluebird": "^3.7.2",
         "cross-fetch": "^4.0.0",
         "crypto-browserify": "^3.12.1",
+        "decimal.js": "^10.4.3",
```
theblockstalk, line 58: dont you need to also add the new SDK version?

src/utils/chain/types.ts
```diff
@@ -182,50 +183,47 @@ export interface IAsset {
...
-        return new Asset(this.token, this.amount + other.getAmount());
+        return new Asset(this.token, this.amount.add(other.getAmount()));
     }
     async getUsdValue(): Promise<number> {
         const price = await this.token.getUsdPrice();
 
         if (price) {
-            const precision = this.token.getPrecision();
-            const precisionMultiplier = BigInt(10) ** BigInt(precision);
+            const precisionMultiplier = new Decimal(10).pow(this.token.getPrecision());
```
theblockstalk, line 202: this looks wrong. we should not have to use precision, just the two Decimal values multiplied by each other. need to make sure that the decimal number, when stored in Asset class is the decimal number, not the integer value. aka for 0.001 ETH we should store Decimal("0.0001") not Decimal("1000000000000000") the Wei value. then you can just do the multiplication and not worry about the precision

theblockstalk: same on L222

PR #1396, opened 2025-02-04
Bug/1301  update sso login response logic

sadiabbasi:


src/containers/SSOLoginContainer.tsx
```diff
@@ -88,12 +89,35 @@ export default function SSOLoginContainer({ payload, platform }: { payload: stri
             if (!responsesManager) throw new Error('Responses manager is not set');
 
             await responsesManager.createResponses(user);
+            //try catch , u
```
theblockstalk, line 92: remove comment. I would add a comment that explains why we have the below logic due to android deeplink vs iOS universal link

src/containers/SSOLoginContainer.tsx
```diff
@@ -88,12 +89,35 @@ export default function SSOLoginContainer({ payload, platform }: { payload: stri
...
+                if (e instanceof SdkError && e.code === SdkErrors.ResponsesNotFound) {
+                    callbackUrl = await user.acceptLoginRequest(responsesManager, platform, {
+                        callbackPath: responsesManager.getExternalLoginResponseOrThrow().getResponse().getPayload()
+                            .callbackPath,
+                        callbackOrigin: responsesManager.getExternalLoginResponseOrThrow().getResponse().getPayload()
+                            .origin,
+                        messageRecipient: responsesManager.getExternalLoginResponseOrThrow().getRequest().getIssuer(),
+                    });
+                } else {
+                    errorStore.setError({
```
theblockstalk, line 113: just throw the error. it will get caught and show to the user in a lower try/catch

PR #1410, opened 2025-02-11
Feature/1322 confirm staking

sadiabbasi:


src/containers/ConfirmStakingContainer.tsx
```diff
@@ -0,0 +1,114 @@
...
+        }
+    };
+
+    return (
+        <>
+            <View style={styles.container}>
+                <TH1 style={styles.vestedHead}>
+                    Confirm staking {formatTokenValue(new Decimal(amount))} {chain.getNativeToken().getSymbol()}
+                </TH1>
+                <TP style={styles.vestedSubHead}>These coins will be locked for 14 days</TP>
```
theblockstalk, line 59: {{14 days}} get this from the staking contract so it doesnt need to be changed if the timeframe changes. also in all other places

src/containers/ConfirmUnStakingContainer.tsx
```diff
@@ -0,0 +1,107 @@
...
+        }
+    };
+
+    return (
+        <>
+            <View style={styles.container}>
+                <TH1 style={styles.vestedHead}>
+                    Confirm unstaking {formatTokenValue(new Decimal(amount))} {chain.getNativeToken().getSymbol()}
+                </TH1>
+                <TP style={styles.vestedSubHead}>Your LEOS will enter a 5-day release period with no yield</TP>
```
theblockstalk, line 52: use the staking contract release period value

src/containers/StakeAssetDetailContainer.tsx
```diff
@@ -0,0 +1,334 @@
...
+    chain: IChain;
+    loading?: boolean;
+};
+
+export function getUnlockStatus(allocation) {
+    const now = new Date();
+    const unstakeableTime = new Date(allocation.unstakeableTime);
+    const releaseTime = new Date(allocation.releaseTime);
+
+    const daysUntilUnlockable = Math.ceil((unstakeableTime.getTime() - now.getTime()) / (1000 * 60 * 60 * 24));
```
theblockstalk, line 30: better to use constant values. we have time.ts helpers with these either in Tonomy ID, or in the SDK or both

src/containers/StakeAssetDetailContainer.tsx
```diff
@@ -0,0 +1,334 @@
...
+    const onClose = () => {
+        refMessage.current?.close();
+    };
+
+    useFocusEffect(
+        useCallback(() => {
+            if (propsLoading) {
+                setLoading(true);
+                setTimeout(() => {
+                    setLoading(false);
```
theblockstalk, line 83: why have the timeout?

src/utils/chain/antelope.ts
```diff
@@ -38,18 +38,32 @@ import {
...
 import { captureError } from '../sentry';
 import { AntelopePushTransactionError, HttpError } from '@tonomy/tonomy-id-sdk';
 import Decimal from 'decimal.js';
+import RNKeyManager from '../RNKeyManager';
 
 const vestingContract = VestingContract.Instance;
+const stakingContract = StakingContract.Instance;
+const eosioTokenContract = EosioTokenContract.Instance;
+
+function getAccountSigner() {
```
theblockstalk, line 64: as this signer is used a couple of times in the app, it is probably worth having this signer exported from the user object instead.

src/utils/chain/antelope.ts
```diff
@@ -405,6 +421,22 @@ export class AntelopeToken extends AbstractToken implements IToken {
...
+    async stakeTokens(account: IAccount, amount: string): Promise<PushTransactionResponse> {
+        throw new Error(`stakeTokens() method not implemented' ${account} ${amount}`);
+    }
+
+    async unStakeTokens(account: IAccount, allocationId: number): Promise<PushTransactionResponse> {
+        throw new Error(`unStakeTokens() method not implemented' ${account} ${allocationId}`);
+    }
 }
 
 export class PangeaVestedToken extends AntelopeToken {
```
theblockstalk, line 442: rename class to just "PangeaToken" as its also about staking, not just vesting

src/utils/chain/antelope.ts
```diff
@@ -436,6 +468,30 @@ export class PangeaVestedToken extends AntelopeToken {
...
+        const accountSigner = getAccountSigner();
+
+        return await stakingContract.stakeTokens(account.getName(), amount, accountSigner);
+    }
+
+    async unStakeTokens(account: IAccount, allocationId: number): Promise<PushTransactionResponse> {
+        const accountSigner = getAccountSigner();
+
+        return await stakingContract.requestUnstake(account.getName(), allocationId, accountSigner);
+    }
```
theblockstalk, line 494: what about releasing tokens. when is this done in the app?

src/utils/chain/etherum.ts
```diff
@@ -218,6 +220,22 @@ export class EthereumToken extends AbstractToken {
     async getAvailableBalance(account?: IAccount): Promise<IAsset> {
         return this.getBalance(account);
     }
+
+    async withdrawVestedTokens(account: IAccount): Promise<PushTransactionResponse> {
```
theblockstalk, line 224: you are better off putting all this in `AbstractToken` instead of both the Antelope and Ethereum tokens as it is the same.

src/utils/chain/types.ts
```diff
@@ -267,6 +268,10 @@ export interface IToken {
     getVestedTokens(account: IAccount): Promise<VestedTokens>;
     getAvailableBalance(account?: IAccount): Promise<IAsset>;
     getVestedTotalBalance(account?: IAccount): Promise<IAsset>;
```
theblockstalk, line 270: probably want to add a function `isStakeable(): boolean;`

src/utils/setSettings.ts
```diff
@@ -2,10 +2,12 @@ import { setSettings } from '@tonomy/tonomy-id-sdk';
...
 
 setSettings({
+    environment: settings.env,
     blockchainUrl: settings.config.blockchainUrl,
     accountSuffix: settings.config.accountSuffix,
     communicationUrl: settings.config.communicationUrl,
     tonomyIdSchema: settings.config.tonomyIdSlug,
     accountsServiceUrl: settings.config.accountsServiceUrl,
     ssoWebsiteOrigin: settings.config.ssoWebsiteOrigin,
+    currencySymbol: 'LEOS',
```
theblockstalk, line 12: better add this to the settings config instead

src/utils/time.ts
```diff
@@ -9,3 +9,10 @@ export function formatDate(date: Date): string {
 export function sleep(milliseconds: number) {
     return new Promise((resolve) => setTimeout(resolve, milliseconds));
 }
+
+export const getStakeUntilDate = () => {
+    const date = new Date();
+
+    date.setDate(date.getDate() + 14);
```
theblockstalk, line 16: use the staking contract constant instead of 14

src/containers/StakeLeosContainer.tsx
theblockstalk, line 1: better call it the StakeAssetContainer, keep it consistent

src/containers/StakeLeosContainer.tsx
```diff
@@ -0,0 +1,287 @@
...
+                const tokenEntry = await getTokenEntryByChain(chain);
+                const account = await getAccountFromChain(tokenEntry, user);
+                const state = await token.getAccountStateData(account);
+
+                setStakingState(state);
+
+                if (state.allocations.length > 0) {
+                    const unstakeDate = state.allocations[0].unstakeableTime;
+                    const today = new Date();
+                    const diffDays = Math.ceil((unstakeDate.getTime() - today.getTime()) / (1000 * 60 * 60 * 24));
```
theblockstalk, line 67: use time constants instead.

src/containers/StakeLeosContainer.tsx
```diff
@@ -0,0 +1,287 @@
...
+    const handleAmountChange = async (input: string) => {
+        setAmount(input);
+        const numericAmount = parseFloat(input) || 0;
+
+        let errorMessage: string | null = null;
+
+        if (numericAmount > parseFloat(availableBalance)) {
+            errorMessage = 'Not enough balance';
+        } else if (numericAmount < minimumStakeTransfer) {
+            errorMessage = `Minimum stake: ${minimumStakeTransfer.toFixed(3)} LEOS`;
```
theblockstalk, line 92: show with 4 decimal places. i think Pavel may not have designs all updated. good to ask him if there are any discrepancies.

src/containers/StakeLeosContainer.tsx
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        setAmountError(errorMessage);
+
+        const calculatedYield = numericAmount * (Math.pow(1 + StakingContract.MAX_APY, 1 / 12) - 1);
```
theblockstalk, line 97: better to make a modular function in the staking contract so that it shows why this formular is called, which can be called here and also in the staking contract. `calculateYield(amount: number, apy: number): Promise<number>`

src/containers/StakingManagerContainer.tsx
theblockstalk, line 1: should this container/screen name be updated? its not the staking manager but the asset manager. im a bit confused as it does vesting but not staking

src/utils/chain/antelope.ts
```diff
@@ -415,7 +447,7 @@ export class PangeaVestedToken extends AntelopeToken {
         return availableBalance.add(vestedBalance);
```
theblockstalk, line 447: this should also be adding the staked coins to the total balance, right?

src/containers/UnstakeAssetContainer.tsx
```diff
@@ -0,0 +1,148 @@
...
+                        <TextInput
+                            style={styles.input}
+                            placeholder="Enter amount"
+                            placeholderTextColor={theme.colors.tabGray}
+                        />
+                        <TouchableOpacity style={styles.inputButton}>
+                            <Text style={styles.inputButtonText}>MAX</Text>
+                        </TouchableOpacity>
+                    </View>
+                    <Text style={styles.inputHelp}>Unlockable: 50,000.00 LEOS</Text>
```
theblockstalk, line 29: i dont think this should be hardcoded...

src/containers/UnstakeAssetContainer.tsx
```diff
@@ -0,0 +1,148 @@
...
+                </View>
+            </View>
+            <View style={styles.unlockAssetView}>
+                <Text style={styles.unlockhead}>Unstacking your coins </Text>
+                <Text style={styles.lockedParagraph}>
+                    Assets will be locked for 5 days during release, during which no interest will be earned
```
theblockstalk, line 35: use the staking contract constant

src/containers/VestedAssetsContainer.tsx
```diff
@@ -3,50 +3,61 @@ import { VestedAssetscreenNavigationProp } from '../screens/VestedAssetsScreen';
...
-            const account = await getAccountFromChain(tokenEntry, user);
-            const allocations = await token.getVestedTokens(account);
-            const usdPriceValue = await chain.getNativeToken().getUsdPrice();
+                const account = await getAccountFromChain(tokenEntry, user);
 
-            setUsdPrice(usdPriceValue);
-            setVestedAllocation(allocations);
-            setLoading(false);
+                const allocations = await token.getVestedTokens(account);
+                const usdPriceValue = await chain.getNativeToken().getUsdPrice();
```
theblockstalk, line 48: it might be worth adding a buffered timeout on this value, aka it only calls the API if > 30 seconds have elapsed, otherwise it uses a cached value. otherwise it calls the same API every screen for almost the same value returned.

src/containers/VestedAssetsContainer.tsx
```diff
@@ -70,6 +81,17 @@ const VestedAssetsContainer = ({ navigation, chain }: VestedAssetProps) => {
         return totalLockedAndUnlockable > 0 ? totalWeightedMultiplier / totalLockedAndUnlockable : 0;
     };
 
+    useFocusEffect(
+        useCallback(() => {
+            if (propsLoading) {
+                setLoading(true);
+                setTimeout(() => {
+                    setLoading(false);
```
theblockstalk, line 89: why set a timeout?

src/containers/WithDrawVestedContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+    amount: number;
+    total: number;
+};
+
+const WithDrawVestedContainer = ({ navigation, chain, amount, total }: VestedAssetSuccessProps) => {
+    const [loading, setLoading] = useState(false);
+    const errorStore = useErrorStore();
+    const [usdPrice, setUsdPrice] = useState(0);
+    const [stakingValues, setStakingValues] = useState({
+        apy: 0,
```
theblockstalk, line 28: might be better off setting this from the contract constant instead of 0

src/containers/WithDrawVestedContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+                const account = await getAccountFromChain(tokenEntry, user);
+                const usdPriceValue = await chain.getNativeToken().getUsdPrice();
+
+                setUsdPrice(usdPriceValue);
+                const accountData = await token.getAccountStateData(account);
+
+                setStakingValues({
+                    apy: accountData.settings.apy,
+                    monthlyEarnings:
+                        formatTokenValue(new Decimal(amount * (Math.pow(1 + StakingContract.MAX_APY, 1 / 12) - 1))) +
```
theblockstalk, line 68: here use the helper function I suggested before.

src/containers/WithDrawVestedContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+                        ' LEOS',
+                });
+            } catch (e) {
+                if (e.message === 'Account not found in staking contract') {
```
theblockstalk, line 72: try use an SdkError code instead

theblockstalk (commented):
great work here.

Important: where are you calling the releasetoken() function in the smart contract so that LEOS finishes releasing?

src/containers/StakingManagerContainer.tsx
sadiabbasi: not needed i removed this...

src/containers/UnstakeAssetContainer.tsx
sadiabbasi: this is also not used old UI flow... going to delete this

package.json
```diff
@@ -136,7 +136,8 @@
     },
     "resolutions": {
         "jsonld": "link:./node_modules/@digitalcredentials/jsonld",
-        "isomorphic-webcrypto": "link:./node_modules/@sphereon/isomorphic-webcrypto"
+        "isomorphic-webcrypto": "link:./node_modules/@sphereon/isomorphic-webcrypto",
+        "@tonomy/tonomy-id-sdk": "portal:/home/sadia/TonomyFoundation/january/Tonomy-ID-Integration/Tonomy-ID-SDK"
```
theblockstalk, line 140: remove

src/utils/chain/common.ts
```diff
@@ -1,7 +1,26 @@
...
+        return cachedPrice;
+    }
+
+    // Fetch new price from CoinGecko
     const res = await fetch(
         `https://api.coingecko.com/api/v3/simple/price?ids=${token}&vs_currencies=${currency}`
     ).then((res) => res.json());
 
-    return res?.ethereum?.usd;
+    const price = res?.ethereum?.usd;
```
theblockstalk, line 18: better type safety to throw an error if the object is not expected, instead of using nullifier accessors

PR #1419, opened 2025-03-02
feat: half the FDV

theblockstalk:


theblockstalk:
not merging as it uses the `feat` conventional commit and will fail

PR #1482, opened 2025-05-20
feat: :zap: feature/1456 created typeoram app storage 

sadiabbasi:


src/utils/StorageManager/entities/appStorage.ts
```diff
@@ -0,0 +1,34 @@
...
+export class AppStorage {
+    @PrimaryColumn({
+        unique: true,
+        type: 'varchar',
+    })
+    @Index()
+    origin!: string;
+
+    @Column({ type: 'varchar' })
+    accountName!: string;
```
theblockstalk, line 15: should also be unique

theblockstalk: accountName should be the primary key. origin should also be indexed

src/utils/StorageManager/entities/appStorage.ts
```diff
@@ -0,0 +1,31 @@
...
+    origin!: string;
+
+    @Column({ type: 'bool' })
+    isLoggedIn!: boolean;
+
+    @Column({ type: 'bool' })
+    isDataShared!: boolean;
+
+    @Column({ type: 'simple-json' })
+    dataShared?: {
```
theblockstalk, line 21: i wonder also, what will happen if in the future we want to add new data shared items, or remove the previous ones? will we need a migration or can the `simple-json` type handle this? if the `simple-json` can handle it then it will be worth having a version tag so we know what schema to expect

sadiabbasi:
not needed we move this storage to sdk

PR #1527, opened 2025-07-02
Feature/1515 integrate sdk

sadiabbasi:


src/containers/VeriffDataSharingContainer.tsx
```diff
@@ -0,0 +1,343 @@
...
+                            ) : (
+                                <NavArrowUp width={16} height={16} color={theme.colors.primary} />
+                            )}
+                        </View>
+                    </TouchableOpacity>
+
+                    {!personalCollapsed && (
+                        <View style={styles.sectionContent}>
+                            <View style={styles.row}>
+                                <Text style={styles.rowLabel}>Date of birth</Text>
```
sadiabbasi, line 245: In personal data which data we should show? https://www.figma.com/design/cvV48t0f7O2znT6QBxK0Zj/Tonomy-ID-app?node-id=13847-13980&t=kVxoyLHGFnnQNsVB-4

src/utils/StorageManager/repositories/identityVerificationStorageManager.ts
```diff
@@ -0,0 +1,174 @@
+import { IdentityVerificationStorageRepository } from './identityVerificationStorageRepository';
```
sadiabbasi, line 1: we don't need the identityStorage files here right?

theblockstalk: correct, you can remove them all

PR #1568, opened 2025-09-02
Master release: Update android sdk version

sadiabbasi:


.github/workflows/build-apk.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
           sudo du -sh ~/.yarn
 
       - name: Install eas globally
-        run: sudo npm install -g eas-cli@12.6.2
+        run: sudo npm install -g eas-cli@16.3.3
```
theblockstalk, line 78: please update this in other places - i think in the `install_prerequisits.sh` script in the integration repo

App.tsx
```diff
@@ -12,10 +12,44 @@ import Debug from 'debug';
 import { wrap } from './src/utils/sentry';
 import InitializeAppProvider from './src/providers/InitializeApp';
 import { StatusBar } from 'react-native';
+import { BackHandler } from 'react-native';
 
 Debug.enable(process.env.DEBUG);
 console.log('DEBUG:', process.env.DEBUG);
 
+/**
```
theblockstalk, line 20: should be in https://github.com/Tonomy-Foundation/Tonomy-ID/blob/testnet/src/utils/polyfill.ts

package.json
```diff
@@ -1,150 +1,162 @@
...
+  "private": true,
+  "packageManager": "yarn@4.3.1",
+  "engines": {
+    "node": ">=20.0.0",
+    "yarn": "^4.3.1"
+  },
+  "expo": {
+    "doctor": {
+      "reactNativeDirectoryCheck": {
+        "exclude": [
```
theblockstalk, line 153: why are these still excluded? shouldnt this be fixed? this for me is a reason to not approve this PR

prepare.app.config.ts
```diff
@@ -109,6 +109,28 @@ const expo: ExpoConfig = {
...
+                ios: {
+                    // You can also override the shared configurations for iOS
+                    customBuildFlags: ['-DSQLITE_ENABLE_DBSTAT_VTAB=1 -DSQLITE_ENABLE_SNAPSHOT=1'],
+                },
+            },
+        ],
+        [
+            'expo-camera',
+            {
+                cameraPermission: 'Allow tonomy id to access your camera',
```
theblockstalk, line 131: "Tonomy ID"

prepare.app.config.ts
```diff
@@ -109,6 +109,28 @@ const expo: ExpoConfig = {
         favicon: settings.config.images.logo48,
     },
     plugins: [
+        [
+            'expo-sqlite',
```
theblockstalk, line 113: why are we overriding the defaults here?
also I see you have used the web config setup from here https://docs.expo.dev/versions/latest/sdk/sqlite/#web-setup however we are not compiling to web so this does not seem necessary

package.json
sadiabbasi: i added in excluded otherwise it throws the error even the packages are updated to latest may be i can create a new task to replace these packages later:
Untested on New Architecture: iconoir-react-native, react-native-exception-handler
  Unmaintained: react-native-exception-handler, react-native-swiper

PR #1570, opened 2025-09-04
Upgrade expo sdk version to 53

sadiabbasi:


android.manifest.plugin.js
```diff
@@ -29,6 +29,30 @@ const updateAndroidManifest = function (config) {
             return perm;
         });
 
+        // ✅ Add or override the ML Kit dependency meta-data
```
theblockstalk, line 32: why are these changes necessary?

.github/workflows/build-apk.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
           sudo du -sh ~/.yarn
 
       - name: Install eas globally
-        run: sudo npm install -g eas-cli@12.6.2
+        run: sudo npm install -g eas-cli@16.3.3
```
theblockstalk, line 78: do a text search for `eas-cli` and update in other places

metro.config.js
```diff
@@ -31,7 +49,6 @@ const debugModulePath = path.resolve(__dirname, 'src/utils/debug.ts');
 // this should be instead use the ./src/utils/debugAndLog.ts
 // so that we can send the logs to Sentry
 config.resolver.resolveRequest = (context, moduleName, platform) => {
-    // if the module is debug and not importing from the src/utils/debug.ts file
```
theblockstalk, line 34: good to keep this comments IMO

package.json
```diff
@@ -1,150 +1,162 @@
...
+    "@wharfkit/antelope": "1.0.13"
+  },
+  "private": true,
+  "packageManager": "yarn@4.3.1",
+  "engines": {
+    "node": ">=20.0.0",
+    "yarn": "^4.3.1"
+  },
+  "expo": {
+    "doctor": {
```
theblockstalk, line 151: please explain?

src/containers/MainSplashContainer.tsx
```diff
@@ -28,7 +28,6 @@ export default function MainSplashScreenContainer({ navigation }: { navigation:
     useEffect(() => {
         async function main() {
             await sleep(800);
-            logout("Invalid data in user's storage");
```
theblockstalk, line 31: is this correct?

src/navigation/BottomTabNavigator.tsx
```diff
@@ -20,35 +22,36 @@ import AppInstructionProvider from '../providers/AppInstruction';
 
 const Tab = createBottomTabNavigator<BottonNavigatorRouteStackParamList>();
 
-type ScanTabBarButtonProps = BottomTabBarButtonProps;
+const isIpad = Platform.OS === 'ios' && Dimensions.get('window').width >= 768;
```
theblockstalk, line 25: dont we have a helper somewhere that determines what device? i think this should be moved there if it exists, or created there if it does not exist

theblockstalk (commented):
Screenshot 2025-09-09 102947
Screenshot 2025-09-09 103001
these are good comments. I would also like to see this for the `android.manifest.plugin.js` and `App.tsx` and `jest.config.ts` changes because they are also not obvious why you have added them. I'm sure you have a reason

package.json
sadiabbasi: https://docs.expo.dev/versions/latest/config/package-json/ getting error with run the command then i did this solution

sadiabbasi: otherwise it throws this error: ✖ Validate packages against React Native Directory package metadata

The following issues were found when validating your dependencies against React Native Directory:
  Untested on New Architecture: iconoir-react-native, react-native-exception-handler
  Unmaintained: react-native-exception-handler, react-native-swiper

src/containers/MainSplashContainer.tsx
sadiabbasi: yess this was the solution which we resolve for each time logout when app close... i added this for testing few things that time.

package.json
sadiabbasi: it suggest to replace this or add the expo doctor in package.json.... 
may be in future we will replace these packages... but for now keep these like this?

theblockstalk: have you tried fixing this? you should do this first before patching. if patch then you need to leave a note in the codebase

sadiabbasi: i tried to check for the iconoir-react-native, it suggest to change the package i updated the package to latest version..
it's time consuming to change this as we are using so many icons from it.
i will leave a note

